PR #5, opened 2016-05-11
Add Accumulate and LastElement (closes #3)

### LINKED ISSUE #3: Move Accumulate and LastElement from hseeberger/akka-commons-streams here
- [Accumulate](https://github.com/hseeberger/commons-akka/blob/master/commons-akka-stream/src/main/scala/de/heikoseeberger/commons/akka/stream/Accumulate.scala#L48): This stage emits folded values like `scan`, but the first element emitted is not the zero value but the result of applying the given function to the given zero value and the first pushed element.
- [LastElement](https://github.com/hseeberger/commons-akka/blob/master/commons-akka-stream/src/main/scala/de/heikoseeberger/commons/akka/stream/LastElement.scala#L43): This stage materializes to the last element pushed before upstream completion, if any, thereby recovering from any failure. Pushed elements are just passed along.

hseeberger:
- Copy src from hseeberger/commons-akka-stream, add dep on ScalaTest
- Remove unnecessary package object
- Rename package to akka.stream.contrib

### DIFF
--- a/src/main/scala/akka/stream/contrib/Accumulate.scala
+++ b/src/main/scala/akka/stream/contrib/Accumulate.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.japi.function
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+
+/**
+ * This companion defines a factory for [[Accumulate]] instances, see [[Accumulate.apply]].
+ */
+object Accumulate {
+
+  /**
+   * Factory for [[Accumulate]] instances.
+   *
+   * @param zero zero value for folding
+   * @param f binary operation for folding
+   * @tparam A input type
+   * @tparam B output type
+   * @return [[Accumulate]] instance
+   */
+  def apply[A, B](zero: B)(f: (B, A) => B): Accumulate[A, B] = new Accumulate(zero)(f)
+
+  /**
+   * Java API: Factory for [[Accumulate]] instances.
+   *
+   * @param zero zero value for folding
+   * @param f binary operation for folding
+   * @tparam A input type
+   * @tparam B output type
+   * @return [[Accumulate]] instance
+   */
+  def create[A, B](zero: B, f: function.Function2[B, A, B]): Accumulate[A, B] = new Accumulate(zero)(f.apply)
+}
+
+/**
+ * This stage emits folded values like `scan`, but the first element emitted is not the zero value but the result of
+ * applying the given function to the given zero value and the first pushed element.
+ *
+ * @param zero zero value for folding
+ * @param f binary operation for folding
+ * @tparam A input type
+ * @tparam B output type
+ */
+final class Accumulate[A, B] private (zero: B)(f: (B, A) => B) extends GraphStage[FlowShape[A, B]] {
+
+  override val shape = FlowShape(Inlet[A]("accumulate.in"), Outlet[B]("accumulate.out"))
+
+  override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+    import shape._
+
+    private var acc = zero
+
+    setHandler(in, new InHandler {
+      override def onPush() = {
+        acc = f(acc, grab(in))
+        push(out, acc)
+      }
+    })
+
+    setHandler(out, new OutHandler {
+      override def onPull() = pull(in)
+    })
+  }
+}
--- a/src/main/scala/akka/stream/contrib/LastElement.scala
+++ b/src/main/scala/akka/stream/contrib/LastElement.scala
@@ -1,0 +1,76 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler }
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * This companion defines a factory for [[LastElement]] instances, see [[LastElement.apply]].
+ */
+object LastElement {
+
+  /**
+   * Factory for [[LastElement]] instances.
+   *
+   * @tparam A input and output type
+   * @return [[LastElement]] instance
+   */
+  def apply[A](): LastElement[A] = new LastElement
+
+  /**
+   * Java API: Factory for [[LastElement]] instances.
+   *
+   * @tparam A input and output type
+   * @return [[LastElement]] instance
+   */
+  def create[A](): LastElement[A] = new LastElement
+}
+
+/**
+ * This stage materializes to the last element pushed before upstream completion, if any, thereby recovering from any
+ * failure. Pushed elements are just passed along.
+ *
+ * @tparam A input and output type
+ */
+final class LastElement[A] private extends GraphStageWithMaterializedValue[FlowShape[A, A], Future[Option[A]]] {
+
+  override val shape = FlowShape(Inlet[A]("lastElement.in"), Outlet[A]("lastElement.out"))
+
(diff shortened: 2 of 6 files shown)


src/main/scala/akka/stream/contrib/LastElement.scala
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        override def onUpstreamFinish() = {
+          matValue.success(currentElement)
+          super.onUpstreamFinish()
+        }
+
+        override def onUpstreamFailure(t: Throwable) = {
+          matValue.success(currentElement)
+          super.onUpstreamFinish()
```
drewhk: interesting, why does it not fail? Should this be configurable instead?

src/test/scala/akka/stream/contrib/AccumulateSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+  "Accumulate" should {
+    "emit folded vaules starting with the result of applying the given function to the given zero and the first pushed element" in {
+      Source(Vector(1, 2, 3, 4, 5))
+        .via(Accumulate(0)(_ + _))
+        .runWith(TestSink.probe)
+        .request(99)
+        .expectNext(1, 3, 6, 10, 15)
+        .expectComplete()
+    }
+  }
```
drewhk: tests for empty sequence, failure,

drewhk:
Welcome, first contributor! :)

src/main/scala/akka/stream/contrib/LastElement.scala
hseeberger: That's the purpose of this stage, see the ScalaDoc comment. I'm using it in a reconnecting SSE client (in akka-sse).

hseeberger:
@drewhk, I have added more tests for `Accumulate`.

src/main/resources/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
patriknw: we don't have this in any other akka project, I guess it doesn't hurt, but then it should match up with https://github.com/akka/akka-stream-contrib/blob/master/LICENSE
(which should change Typesafe to Lightbend, I see now)

hseeberger: Ah, I forgot to delete that. Will do it now.

hseeberger:
@patriknw, license file deleted.

src/main/scala/akka/stream/contrib/Accumulate.scala
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Factory for [[Accumulate]] instances.
+   *
+   * @param zero zero value for folding
+   * @param f binary operation for folding
+   * @tparam A input type
+   * @tparam B output type
+   * @return [[Accumulate]] instance
+   */
+  def apply[A, B](zero: B)(f: (B, A) => B): Accumulate[A, B] = new Accumulate(zero)(f)
```
patriknw: should it also have a java friendly variant or do we let java users make use of the compat lib themselves?

hseeberger: I don't know, I have no experience with Java APIs. @patriknw, could you please make a suggestion for a Java friendly overload?

src/test/scala/akka/stream/contrib/BaseStreamSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+  protected implicit val system = {
+    def config = ConfigFactory.parseString(s"akka.stream.materializer.auto-fusing=$autoFusing")
+      .withFallback(ConfigFactory.load())
+    ActorSystem("default", config)
+  }
+
+  protected implicit val mat = ActorMaterializer()
+
+  override protected def afterAll() = {
+    Await.ready(system.terminate(), Duration.Inf)
```
patriknw: use some timeout, otherwise there is a risk that CI environment hangs here if there is a problem

patriknw:
thanks for contributing :-)

src/test/scala/akka/stream/contrib/BaseStreamSpec.scala
hseeberger: Really? What happens if the system can't terminate? I think the JVM won't shutdown in any case.

src/main/scala/akka/stream/contrib/Accumulate.scala
patriknw: see `akka.stream.javadsl.Sink#fold` for inspiration, name can be `def create`

src/test/scala/akka/stream/contrib/BaseStreamSpec.scala
patriknw: The jvm will be killed no matter what when all tests are completed, but the problem with this is that Travis doesn't know that this is completed/stalled. Eventually some max build timeout will kick in, but that is unecessary. Just make it `30.seconds`.

src/main/scala/akka/stream/contrib/Accumulate.scala
hseeberger: Why is `akka.japi.function.Function2` needed? Scala 2.12 and Java 8 means that we can use Scala function types in the signature and call the function with Java lambdas as arguments.

src/test/scala/akka/stream/contrib/BaseStreamSpec.scala
hseeberger: OK

hseeberger:
All right, I think I have addressed all of your concerns.

src/main/scala/akka/stream/contrib/Accumulate.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * This stage emits folded values like `scan`, but the first element emitted is not the zero value but the result of
+ * applying the given function to the given zero value and the first pushed element.
+ *
+ * @param zero zero value for folding
+ * @param f binary operation for folding
+ * @tparam A input type
+ * @tparam B output type
+ */
+final class Accumulate[A, B] private (zero: B)(f: (B, A) => B) extends GraphStage[FlowShape[A, B]] {
```
johanandren: Not that this is a lot of code like this, but:
```
def accumulate[A, B](zero: B, f: (B, A) => B): Flow[A, B, NotUsed] = Flow[A].statefulMapConcat { () =>
    var acc = zero

    { (elem: A) =>
      acc = f(acc, elem)
      List(acc)
    }
  }
```

johanandren:
LGTM

src/main/scala/akka/stream/contrib/Accumulate.scala
hseeberger: Thanks @johanandren, but I'm a bit on the fence with allocating a `List` just to push a single element.

johanandren: Good point

src/main/scala/akka/stream/contrib/Accumulate.scala
patriknw: we are not on 2.12 yet, are you?

I think we introduced akka.japi.function.* because of variance (and akka streams were originally compiled with java6)

src/test/scala/akka/stream/contrib/BaseStreamSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+  protected implicit val system = {
+    def config = ConfigFactory.parseString(s"akka.stream.materializer.auto-fusing=$autoFusing")
+      .withFallback(ConfigFactory.load())
+    ActorSystem("default", config)
+  }
+
+  protected implicit val mat = ActorMaterializer()
+
+  override protected def afterAll() = {
+    Await.ready(system.terminate(), 42.seconds)
```
patriknw: :-)

patriknw:
LGTM

PR #9, opened 2016-05-16
A Source for zip files (fixes #8)

### LINKED ISSUE #8: Source for zip files
A stage that works as a `Source` of data chunks extracted from zip files. In addition to regular files, the zip file might contain directories and other zip files.

juanjovazquez:
closes #8

### DIFF
--- a/src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
+++ b/src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
@@ -1,0 +1,226 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import java.util.zip.{ ZipEntry, ZipInputStream }
+import akka.stream.Attributes.{ InputBuffer, name }
+import akka.stream.contrib.ZipInputStreamSource.{ DefaultAllowedZipExtensions, DefaultChunkSize, ZipEntryData }
+import akka.stream.impl.Stages.DefaultAttributes.IODispatcher
+import akka.stream.scaladsl.Source
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, OutHandler }
+import akka.stream.{ Attributes, Outlet, SourceShape }
+import akka.util.ByteString
+import akka.util.ByteString.ByteString1C
+import scala.annotation.tailrec
+import scala.collection.immutable
+import scala.concurrent.{ Future, Promise }
+import scala.util.control.NonFatal
+
+/**
+ * This companion defines a factory for [[ZipInputStreamSource]] instances,
+ * see [[ZipInputStreamSource.apply]].
+ */
+object ZipInputStreamSource {
+
+  final val DefaultChunkSize = 8192
+  final val DefaultAllowedZipExtensions = immutable.Seq(".zip")
+
+  /**
+   * Data type for zip entries.
+   *
+   * @param name file name
+   * @param creationTime The last modification time of the entry in milliseconds
+   *          since the epoch, or -1 if not specified
+   */
+  case class ZipEntryData(name: String, creationTime: Long)
+
+  /**
+   * Factory for [[ZipInputStreamSource]] instances wrapped
+   * into [[Source]].
+   *
+   * @param in a function that builds a [[ZipInputStream]]
+   * @param chunkSize the size of the chunks
+   * @param allowedZipExtensions collection of allowed extensions for zipped containers,
+   *                    as zip or jar files. Only ".zip" extension allowed by default.
+   * @return [[ZipInputStreamSource]] instance
+   */
+  def apply(
+    in:                   () => ZipInputStream,
+    chunkSize:            Int                   = DefaultChunkSize,
+    allowedZipExtensions: immutable.Seq[String] = DefaultAllowedZipExtensions
+  ): Source[(ZipEntryData, ByteString), Future[Long]] =
+    Source.fromGraph(new ZipInputStreamSource(in, chunkSize, allowedZipExtensions))
+      .withAttributes(name("zipInputStreamSource") and IODispatcher)
+
+  /**
+   * Java API: Factory for [[ZipInputStreamSource]] instances wrapped
+   * into [[Source]].
+   *
+   * @param in a function that builds a [[ZipInputStream]]
+   * @param chunkSize the size of the chunks
+   * @param allowedZipExtensions collection of allowed extensions for zipped containers,
+   *                    as zip or jar files. Only ".zip" extension allowed by default.
+   * @return [[ZipInputStreamSource]] instance
+   */
+  def create(
+    in:                   Function0[ZipInputStream],
+    chunkSize:            Int                       = DefaultChunkSize,
+    allowedZipExtensions: immutable.Seq[String]     = DefaultAllowedZipExtensions
+  ): Source[(ZipEntryData, ByteString), Future[Long]] =
+    Source.fromGraph(new ZipInputStreamSource(in, chunkSize, allowedZipExtensions))
+      .withAttributes(name("zipInputStreamSource") and IODispatcher)
+}
+
+/**
+ * A stage that works as a [[Source]] of data chunks extracted from
+ * zip files. In addition to regular files, the zip file might contain
+ * directories and other zip files. Every chunk is a tuple of [[ZipEntryData]]
+ * and [[ByteString]], where the former carries basic info about the file from
+ * which the bytes come and the latter carries those bytes. This stage
+ * materializes to the total amount of read bytes.
+ *
+ * @param in a function that builds a [[ZipInputStream]]
+ * @param chunkSize the size of the chunks
+ */
+final class ZipInputStreamSource private (
+  in:                   () => ZipInputStream,
+  chunkSize:            Int                   = DefaultChunkSize,
+  allowedZipExtensions: immutable.Seq[String] = DefaultAllowedZipExtensions
+) extends GraphStageWithMaterializedValue[SourceShape[(ZipEntryData, ByteString)], Future[Long]] {
+
+  val matValue = Promise[Long]()
+
+  override val shape =
+    SourceShape(Outlet[(ZipEntryData, ByteString)]("zipInputStreamSource.out"))
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
+
+    val InputBuffer(initialBuffer, maxBuffer) =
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @juanjovazquez,

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * This companion defines a factory for [[ZipInputStreamSource]] instances,
+ * see [[ZipInputStreamSource.apply]].
+ */
+object ZipInputStreamSource {
+
+  val DefaultChunkSize: Int = 8192
```
hseeberger: Use a constant value definition (see Scala Language Definition, section 1.4): `final val DefaultChunkSize = 8192`

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Factory for [[ZipInputStreamSource]] instances wrapped
+   * into [[Source]].
+   *
+   * @param in a function that builds a [[ZipInputStream]]
+   * @param chunkSize the size of the chunks
+   * @return [[ZipInputStreamSource]] instance
+   */
+  def apply(
```
hseeberger: This is fancily formatted. Shouldn't Scalariform treat this differently?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
+
+    val InputBuffer(initialBuffer, maxBuffer) =
+      inheritedAttributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16))
+
+    val logic = new GraphStageLogic(shape) {
+      import shape._
+
+      private var is: ZipInputStream = _
```
hseeberger: Don't use the magic underscore here, use `null` instead.

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+      private var currentStreams: Seq[ZipInputStream] = Seq()
+
+      override def preStart(): Unit = {
+        super.preStart()
+        try {
+          is = in()
+          currentStreams = List(is)
+          fillBuffer(initialBuffer)
+        } catch {
+          case ex: Exception =>
```
hseeberger: Shouldn't this only catch `NonFatal`s?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+            case Some(e) if isZipFile(e) =>
+              nextEntry(new ZipInputStream(z) +: streams)
+            case Some(e) if e.isDirectory =>
+              nextEntry(streams)
+            case _ =>
+              (entry, streams)
+          }
+      }
+
+      private def isZipFile(e: ZipEntry) = e.getName.toLowerCase.endsWith(".zip")
```
hseeberger: Hmm ... what about `".jar"` files? They are zipped, too, but have a different ending. Maybe make this configurable via a parameter of type `immutable.Seq[String]` with a default of `".zip"?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+
+      /** BLOCKING I/O READ */
+      private def readChunk() = {
+        def read(arr: Array[Byte]) = currentStreams.headOption.flatMap { stream =>
+          val readBytes = stream.read(arr)
```
hseeberger: Shouldn't you wrap this and following blocking calls in `scala.concurrent.blocking`?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
juanjovazquez: I think it's the consequence of having `alignParameters` set to `true` in the `formattingPreferences`. Just to clarify, would you want to change this?.

src/test/scala/akka/stream/contrib/ZipInputStreamSourceSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+      })
+    }
+  }
+
+  private def withZos(f: ZipOutputStream => Unit) = {
+    val baos = new ByteArrayOutputStream()
+    val zos = new ZipOutputStream(baos)
+    f(zos)
+    zos.finish()
+    zos.close()
```
patriknw: `close()` is very important when using `GZIPInputStream` for avoiding leaking off-heap memory, I'm not sure about `ZipOutputStream`.
Should it be in a finally anyway?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,206 @@
...
+            is.close()
+          } finally {
+            matValue.success(readBytesTotal)
+            complete(out)
+          }
+        }
+
+        override def onDownstreamFinish(): Unit =
+          try {
+            is.close()
```
patriknw: `close()` is very important when using GZIPInputStream for avoiding leaking off-heap memory. Should we add additional `close()` in `postStop`?

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
juanjovazquez: To be honest I'm not familiar with the `scala.concurrent.blocking` idiom. According to [this discussion](http://stackoverflow.com/questions/29068064/scala-concurrent-blocking-what-does-it-actually-do) the real benefits strongly depend on the `ExecutionContext` involved. I've tried to be a good citizen of Akka :) and followed the @ktoso 's ["Zen of Akka"](http://es.slideshare.net/ktoso/zen-of-akka) by assigning the `IODispatcher` thread pool by default in the attributes. I've run a rough search in the Akka codebase but I couldn't find examples of use of this idiom. So, I'm not very comfortable with this change. Could anyone else throw some light on the matter?.

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
juanjovazquez: Good point. Done.

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
juanjovazquez: Done.

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,226 @@
...
+  final val DefaultAllowedZipExtensions = immutable.Seq(".zip")
+
+  /**
+   * Data type for zip entries.
+   *
+   * @param name file name
+   * @param creationTime The last modification time of the entry in milliseconds
+   *          since the epoch, or -1 if not specified
+   */
+  case class ZipEntryData(name: String, creationTime: Long)
```
patriknw: case classes should be marked `final` (enables some superpowers in the pattern matcher)

patriknw:
LGTM

patriknw:
Thanks for contributing, @juanjovazquez

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
juanjovazquez: Right. I'll open an issue to change that. Thanks!.

patriknw: ok, I merged because I didn't think it was very important, but new PRs are welcome (for such small changes it's fine to not create a separate issue)

src/main/scala/akka/stream/contrib/ZipInputStreamSource.scala
```diff
@@ -0,0 +1,226 @@
...
+   * @param chunkSize the size of the chunks
+   * @param allowedZipExtensions collection of allowed extensions for zipped containers,
+   *                    as zip or jar files. Only ".zip" extension allowed by default.
+   * @return [[ZipInputStreamSource]] instance
+   */
+  def apply(
+    in:                   () => ZipInputStream,
+    chunkSize:            Int                   = DefaultChunkSize,
+    allowedZipExtensions: immutable.Seq[String] = DefaultAllowedZipExtensions
+  ): Source[(ZipEntryData, ByteString), Future[Long]] =
```
2beaucoup: Wouldn't `Source[(ZipEntryData, Source[ByteString]), Future[Long]]` scale better for large files?

juanjovazquez: @2beaucoup Do you mean a source where every element is a new `ByteString` source for a particular zip entry?. Could you elaborate a bit more?.

2beaucoup: Yeah, exactly. That way the entries would be streams themself instead of unbounded `ByteString`s.

Take a look at [StreamOfStreams.scala](https://github.com/akka/akka/blob/03395d5739ee371e518331f66dfd84447a8ff9ce/akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala) for inspiration.

juanjovazquez: Interesting. I'll look into it ASAP. Thanks ;)

drewhk: The API is not public yet, though :( We will make it public likely in the next release. (we weren't sure if the API is fully correct at that time, so we delayed opening up, but I think the time has come finally).

2beaucoup: Right @drewhk . Didn't think of that. What exactly is going to be exposed here?

juanjovazquez: @2beaucoup @drewhk Probably it is wiser to wait to have closed this API, right?.

drewhk: @2beaucoup basically the SubSource and SubSink classes which are just wrappers around AsyncCallbacks, but provide a lot of convenience and the familiar push/pull interfaces.

PR #16, opened 2016-06-01
Add AlphabeticFraming

hseeberger:
Using `AlphabeticFraming` one can easily implement a stream randomly shuffling each word (only alphabetic characters) except for the first and last character. An example:

*Lkue Swyekaklr has vanseihd. In his ancbese, the siinster FRSIT OREDR has resin form the ahess of the Erimpe and will not rset until Slkaywekr, the lsat Jedi, has been deorsyted.*

### DIFF
--- a/src/main/scala/akka/stream/contrib/AlphabeticFraming.scala
+++ b/src/main/scala/akka/stream/contrib/AlphabeticFraming.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+
+object AlphabeticFraming {
+
+  final case class Frame(value: String, isAlphabetic: Boolean)
+
+  private final val alphabet = 'A'.to('Z').to[Set]
+
+  def apply(): AlphabeticFraming = new AlphabeticFraming
+
+  private def isAlphabetic(c: Char) = alphabet.contains(c.toUpper)
+}
+
+class AlphabeticFraming private extends GraphStage[FlowShape[String, AlphabeticFraming.Frame]] {
+  import AlphabeticFraming._
+
+  override val shape = FlowShape(Inlet[String]("alphabeticFraming.in"), Outlet[Frame]("alphabeticFraming.out"))
+
+  override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+    import shape._
+
+    private var frame = Option.empty[Frame]
+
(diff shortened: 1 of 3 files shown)


drewhk:
Well, I think this is a bit unfocused even for this repository :)

hseeberger:
@drewhk, all right. Would you nevertheless take a short look at the implementation to check that there's nothing terribly wrong?

patriknw:
it's a fun thing, but I also have a hard time to see the real use case for this

hseeberger:
@patriknw, you got me, I have to admit that the main reason for this PR is to get you review – this might end up in new training materials which might become Lightbend's (might to the power of two).

src/main/scala/akka/stream/contrib/AlphabeticFraming.scala
```diff
@@ -0,0 +1,54 @@
...
+            case _                                                => Vector(Frame(c.toString, isAlphabetic(c)))
+          } // format: ON
+        }
+        emitMultiple(out, frames.init)
+        frame = Some(frames.last)
+      }
+
+      override def onUpstreamFinish() = {
+        emitMultiple(out, frame.iterator)
+        super.onUpstreamFinish()
```
patriknw: is this working?

patriknw:
ok, it's a great example for trainings. I can't see anything wrong, apart from possibly the `onUpstreamFinish`

src/main/scala/akka/stream/contrib/AlphabeticFraming.scala
hseeberger: @patriknw, what exactly are you asking? Isn't calling `super` the usual way to overwrite a concrete method?

hseeberger: @patriknw, what exactly are you asking? Isn't calling `super` the usual way to overwrite a concrete method?

patriknw: I uncertain myself if things are taken care of automatically for emitting the remaining elements and then completing the stage.

drewhk: You can just call `completeStage` there, that keeps track of the emits (and that is exactly what the super call would do anyway).

hseeberger: @drewhk, well, you know that, but how would I or some other innocent user know which exact method needs to be called – I'd rather stick to call `super`. This can't be wrong, right?

hseeberger: @drewhk, so is there anything wrong with calling `super`?

drewhk: No, you can keep it if you prefer.

hseeberger:
Closing, because out of scope. Thanks for the reviews!

PR #17, opened 2016-06-01
Timed sorting with sortWithIn operation #15 for review

### LINKED ISSUE #15: streams: request new feature: timed sorting
Clone of https://github.com/akka/akka/issues/20643

Can we have timed sorting operation in akka streams?

Similar to `takeWithin`, This could be `sortWithin`

Possible method signatures
```scala
def sortWithin(d: FiniteDuration): Source[Out, Mat]
def sortWithin(d: FiniteDuration, c: Comparator): Source[Out, Mat]
```

anandshah123:
Timed sorting with sortWithIn operation (Closes #15)

### DIFF
--- a/src/main/scala/akka/stream/contrib/SortWithin.scala
+++ b/src/main/scala/akka/stream/contrib/SortWithin.scala
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.Attributes
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.stage.{GraphStageLogic, InHandler, OutHandler, TimerGraphStageLogic}
+
+import scala.collection.immutable
+import scala.concurrent.duration.FiniteDuration
+
+object SortWithin {
+
+  /**
+    * Factory for [[SortWithin]] instances.
+    *
+    * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
+    * @tparam A input type
+    * @return [[SortWithin]] instance.
+    */
+  def apply[A <: Comparable[A]](finiteDuration: FiniteDuration): SortWithin[A] = new SortWithin[A](finiteDuration)
+
+  /**
+    * Java API: Factory for [[SortWithin]] instances.
+    *
+    * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
+    * @tparam A input type
+    * @return [[SortWithin]] instance.
+    */
+  def create[A <: Comparable[A]](finiteDuration: FiniteDuration): SortWithin[A] = new SortWithin[A](finiteDuration)
+}
+
+final class SortWithin[T <: Comparable[T]](timeout: FiniteDuration) extends SimpleLinearGraphStage[T] {
+
+  var sortedElements = immutable.TreeSet[T]()
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        val element: T = grab(in)
+        sortedElements += element
+        pull(in)
+      }
+
+      @scala.throws[Exception](classOf[Exception])
+      override def onUpstreamFinish(): Unit = {
+        emitMultiple(out, sortedElements.iterator)
+        super.onUpstreamFinish()
+      }
+
+      @scala.throws[Exception](classOf[Exception])
+      override def onUpstreamFailure(ex: Throwable): Unit = {
+        emitMultiple(out, sortedElements.iterator)
+        super.onUpstreamFailure(ex)
+      }
+    })
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+
+
+    final override protected def onTimer(key: Any): Unit = {
+      emitMultiple(out, sortedElements.iterator)
+      completeStage()
+    }
+
+    override def preStart(): Unit = scheduleOnce("SortWithinTimer", timeout)
+  }
+
+  override def toString = "SortWithin"
+}
--- a/src/main/scala/akka/stream/contrib/SortWithin.scala
+++ b/src/main/scala/akka/stream/contrib/SortWithin.scala
@@ -7,13 +7,21 @@ import akka.stream.Attributes
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 import akka.stream.stage.{GraphStageLogic, InHandler, OutHandler, TimerGraphStageLogic}
 
-import scala.collection.immutable
-import scala.concurrent.duration.FiniteDuration
+import scala.collection.immutable.VectorBuilder
+import scala.concurrent.duration.{Duration, FiniteDuration}
 
 object SortWithin {
 
   /**
-    * Factory for [[SortWithin]] instances.
+    * Sorts elements received within given duration and emits them at the end of each duration.
+    *
+    * '''Emits when''' at regular interval of finite duration timeout
+    *
+    * '''Backpressures when''' downstream backpressures
+    *
+    * '''Completes when''' upstream completes
+    *
+    * '''Cancels when''' downstream cancels
     *
     * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
     * @tparam A input type
@@ -22,7 +30,15 @@ object SortWithin {
   def apply[A <: Comparable[A]](finiteDuration: FiniteDuration): SortWithin[A] = new SortWithin[A](finiteDuration)
 
   /**
-    * Java API: Factory for [[SortWithin]] instances.
+    * Java API: Sorts elements received within given duration and emits them at the end of each duration.
+    *
+    * '''Emits when''' at regular interval of finite duration timeout
+    *
+    * '''Backpressures when''' downstream backpressures
+    *
+    * '''Completes when''' upstream completes
+    *
+    * '''Cancels when''' downstream cancels
     *
     * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
     * @tparam A input type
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @anandshah123,

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,73 @@
...
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.stage.{GraphStageLogic, InHandler, OutHandler, TimerGraphStageLogic}
+
+import scala.collection.immutable
+import scala.concurrent.duration.FiniteDuration
+
+object SortWithin {
+
+  /**
+    * Factory for [[SortWithin]] instances.
```
ktoso: Please use the usual formatting (as Akka does):

```
/**
   * Factory for [[SortWithin]] instances.
```

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,73 @@
...
+    *
+    * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
+    * @tparam A input type
+    * @return [[SortWithin]] instance.
```
ktoso: no need to document the obvious things, please remove

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,73 @@
...
+    */
+  def apply[A <: Comparable[A]](finiteDuration: FiniteDuration): SortWithin[A] = new SortWithin[A](finiteDuration)
+
+  /**
+    * Java API: Factory for [[SortWithin]] instances.
```
ktoso: good

ktoso: There should be documentation what SoprtWithin actually does, not that we're a factory here :)

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,73 @@
...
+    *
+    * @param finiteDuration duration within which sorting needs to be done. Ideally should be kept as small as possible
+    * @tparam A input type
+    * @return [[SortWithin]] instance.
+    */
+  def create[A <: Comparable[A]](finiteDuration: FiniteDuration): SortWithin[A] = new SortWithin[A](finiteDuration)
+}
+
+final class SortWithin[T <: Comparable[T]](timeout: FiniteDuration) extends SimpleLinearGraphStage[T] {
```
ktoso: what if `0.millis`?

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,73 @@
...
+        pull(in)
+      }
+
+      @scala.throws[Exception](classOf[Exception])
+      override def onUpstreamFinish(): Unit = {
+        emitMultiple(out, sortedElements.iterator)
+        super.onUpstreamFinish()
+      }
+
+      @scala.throws[Exception](classOf[Exception])
```
ktoso: no need to include these annotations, please remove

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      source.sendNext(1)
+      source.sendNext(3)
+      source.sendNext(2)
+      source.sendComplete()
+      sink.expectNext(1, 2, 3)
+      sink.expectComplete()
+    }
+  }
+
+  "A stream via SortWithin" should {
```
ktoso: please nest the 3 tests in the same `"A stream via SortWithin" should {` block :)

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+        .toMat(TestSink.probe)(Keep.both)
+        .run()
+      sink.request(99)
+      source.sendNext(3)
+      source.sendNext(1)
+      val cause: RuntimeException = new RuntimeException
+      source.sendError(cause)
+      sink.expectNext(FiniteDuration.apply(3, TimeUnit.SECONDS), 1)
+      sink.expectError(cause)
+    }
```
ktoso: seems to be missing a test for the case when we emit 3 elements, timer triggers, then we emit some more elements, they should be independently sorted

ktoso: Hm I see this is doing something a bit different...
May I ask on the exact use case?

I was expecting this to "sort as many elements as you can within T time, emit it as sorted sequence".
Why is this like a "take within, but sorted"? If that's the goal one can simply takeWithin and then sort once.

ktoso:
Thanks for the contrib, had some questions in line, most interestingly for me is the actual use case. 
This doesn't seem to do more than takeWithin+sort though, or am I missing something?

src/main/scala/akka/stream/contrib/SortWithin.scala
anandshah123: Added validation for the same

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
anandshah123: Yes you are correct use case is "sort as many elements possible within T time, emit sorted sequence & repeat"

anandshah123:
@ktoso ready for review again

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.Attributes
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
```
drewhk: Please don't use internal classes, use an ordinary GraphStage. (or alternatively, add a LinearGraphStage helper to contrib itself that is no longer internal)

src/main/scala/akka/stream/contrib/SortWithin.scala
drewhk: Does this logic need a GraphStage at all? Can't you just use:

```scala
Flow[A].groupedWithin(duration).mapConcat(identity)
```

anandshah123:
I don't think tarvis build should fail. Can anyone please check?

anandshah123:
@drewhk ready for review

src/main/scala/akka/stream/contrib/SortWithin.scala
```diff
@@ -0,0 +1,97 @@
...
+    })
+
+
+    final override protected def onTimer(key: Any): Unit = {
+      emitMultiple(out, buf.result().sorted)
+      buf.clear()
+    }
+
+    override def preStart(): Unit = {
+      schedulePeriodically("SortWithinTimer", timeout)
```
patriknw: `private case object SortWithinTimer` in the companion would be better than string key

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+    "sort the elements published till given FiniteDuration" in {
+      val (source, sink) = TestSource.probe[Integer]
+        .via(SortWithin[Integer](FiniteDuration.apply(5, TimeUnit.SECONDS)))
+        .toMat(TestSink.probe)(Keep.both)
+        .run()
+      sink.request(99)
+      source.sendNext(1)
+      source.sendNext(3)
+      source.sendNext(2)
+      sink.expectNext(FiniteDuration.apply(6, TimeUnit.SECONDS), 1)
```
patriknw: `import scala.concurrent.duration._`

and then `6.seconds`

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+      sink.expectNext(FiniteDuration.apply(6, TimeUnit.SECONDS), 2)
+      sink.expectNext(FiniteDuration.apply(6, TimeUnit.SECONDS), 3)
```
patriknw: can we speed up this test somewhat?

anandshah123:
@patriknw ready for review

src/test/scala/akka/stream/contrib/SortWithinSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+class SortWithinSpecAutoFusingOff extends {
+  val autoFusing = false
+} with SortWithinSpec
+
+trait SortWithinSpec extends BaseStreamSpec {
+
+  "A stream via SortWithin" should {
+    "sort the elements published till given FiniteDuration" in {
+      val (source, sink) = TestSource.probe[Integer]
+        .via(SortWithin[Integer](500.millis))
```
patriknw: This is a difficult balance, we are running CI on Travis, and that is very unpredictable with timing. Expect that you can have a GC (or other pause) for 2 seconds at any time. Perhaps something in the middle of this and your original values?

anandshah123:
@patriknw done

patriknw:
@anandshah123 did you answer @ktoso and @drewhk questions about why it must be a `GraphStage` instead of using `takeWithin` or `groupWithin`?

anandshah123:
@patriknw Yes. because the intention of having this operation in place is to pass sorted elements to downstream within continuous repeating intervals. Whereas takeWithin & groupWithin are terminal operations.

patriknw:
@drewhk @ktoso are you fine with merging this?

drewhk:
Hm, groupedWithin is not a terminal operation, it will repeatedly emit groups sampled by time window (and possibly capped to a maximum size).

anandshah123:
@drewhk sortWithin will sort elements & emit them to downstream by time window

drewhk:
I still don't understand how is it different from `groupedWithin().mapConcat(identity)`

drewhk:
To quote the docs of `groupedWithin()`:

> Chunk up this stream into groups of elements received within a time window, or limited by the given number of elements, whatever happens first. Empty groups will not be emitted if no elements are received from upstream. The last group before end-of-stream will contain the buffered elements since the previously emitted group.

anandshah123:
@drewhk Yes `groupedWithin().mapConcat(sortFunction)` will produce same output. `sortWithin` will be like shorthand doing the same.

Issue #19, opened 2016-06-05
Add PartitionWith stage


hochgi:
a rather simple, but useful stage:

```scala
class PartitionWith[In,Out0,Out1](p: In => Either[Out0,Out1]) extends GraphStage[FanOutShape2[In,Out0,Out1]]
```

useful when you need partition + map.
instead of having 1 partition + 2 map stages, code can be much simplified with this stage.

PR #20, opened 2016-06-05
PartitionWith stage added

hochgi:
as defined in #19

### DIFF
--- a/src/main/scala/akka/stream/contrib/PartitionWith.scala
+++ b/src/main/scala/akka/stream/contrib/PartitionWith.scala
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.japi.function
+import akka.stream.{ Attributes, FanOutShape2 }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+
+/**
+ * This companion defines a factory for [[PartitionWith]] instances, see [[PartitionWith.apply]].
+ */
+object PartitionWith {
+
+  /**
+   * Factory for [[PartitionWith]] instances.
+   *
+   * @param p partition function
+   * @tparam In input type
+   * @tparam Out0 left output type
+   * @tparam Out1 right output type
+   * @return [[PartitionWith]] instance
+   */
+  def apply[In, Out0, Out1](p: In => Either[Out0, Out1]): PartitionWith[In, Out0, Out1] = new PartitionWith(p)
+
+  /**
+   * Java API: Factory for [[PartitionWith]] instances.
+   *
+   * @param p partition function
+   * @tparam In input type
+   * @tparam Out0 left output type
+   * @tparam Out1 right output type
+   * @return [[PartitionWith]] instance
+   */
+  def create[In, Out0, Out1](p: function.Function[In, Either[Out0, Out1]]): PartitionWith[In, Out0, Out1] = new PartitionWith(p.apply)
+}
+
+/**
+ * This stage partitions input to 2 different outlets,
+ * applying different transformations on the elements,
+ * according to the received partition function.
+ *
+ * @param p partition function
+ * @tparam In input type
+ * @tparam Out0 left output type
+ * @tparam Out1 right output type
+ */
+final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1]) extends GraphStage[FanOutShape2[In, Out0, Out1]] {
+
+  override val shape = new FanOutShape2[In, Out0, Out1]("partitionWith")
+
+  override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+    import shape._
+
+    private var pending: Either[Out0, Out1] = null
+
+    setHandler(in, new InHandler {
+      override def onPush() = {
+        val elem = grab(in)
+        p(elem) match {
+          case Left(o) if isAvailable(out0) =>
+            push(out0, o)
+            if (isAvailable(out1))
+              pull(in)
+          case Right(o) if isAvailable(out1) =>
+            push(out1, o)
+            if (isAvailable(out0))
+              pull(in)
+          case either =>
+            pending = either
+        }
+      }
+
+      override def onUpstreamFinish() = {
+        if (pending eq null)
+          completeStage()
+      }
+    })
+
+    setHandler(out0, new OutHandler {
+      override def onPull() = if (pending ne null) pending.left.foreach { o =>
+        push(out0, o)
+        if (isClosed(in)) completeStage()
+        else {
+          pending = null
+          if (isAvailable(out1))
+            pull(in)
+        }
+      }
+      else if (!hasBeenPulled(in)) pull(in)
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull() = if (pending ne null) pending.right.foreach { o =>
+        push(out1, o)
+        if (isClosed(in)) completeStage()
+        else {
+          pending = null
+          if (isAvailable(out0))
+            pull(in)
+        }
+      }
+      else if (!hasBeenPulled(in)) pull(in)
+    })
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ target/
 *.sublime-project
 /bin/
 ext-lib-src/
+.ensime
+.ensime_cache/
--- a/src/test/scala/akka/stream/contrib/PartitionWithSpec.scala
+++ b/src/test/scala/akka/stream/contrib/PartitionWithSpec.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.scaladsl._
+import akka.stream.{ FlowShape }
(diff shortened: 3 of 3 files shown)


drewhk:
Does this really need a new stage? Can't this be simply created by using the generic partition graphstage and rewrapping it in a graph of the desired fanout shape?

ktoso:
+1 for implementing in terms of known stages

hochgi:
@drewhk @ktoso  IMO it is. many stages can be expressed in terms of other simpler ones. I find this one extremely useful. of course it can be expressed more verbosely with a partition stage and 2 map stages for each out. it's a bit verbose, and I like the elegance of this.

BTW, I also added `partitionWith` as a combinator on regular collections in our code base, and it seems to caught on.

ktoso:
What we mean is that your implementation of this logic does not have to be a `GraphStage`, just build it using a Graph and expose it - the "more verbose way" is hidden away internally, users don't see it, and we don't have to maintain a full custom graph stage impl to achieve the same thing

hochgi:
@ktoso you mean something like:
```scala
def partitionWith[In,Out0,Out1](f: In => Either[Out0,Out1]) = GraphDSL.create() { implicit b =>
  import GraphDSL.Implicits._

  val map0 = b.add(Flow[In].map(f andThen {_.left.get}))
  val map1 = b.add(Flow[In].map(f andThen {_.right.get}))
  val part = b.add(Partition[In](2,{ i =>
    f(i).fold(_ => 0,_ => 1)
  }))

  part.out(0) ~> map0
  part.out(1) ~> map1
  new FanOutShape2(part.in, map0.out, map1.out)
}
```

drewhk:
Yes, exactly.

hochgi:
@drewhk @ktoso OK, I see.
but what about performance?

here's a quick(?) benchmarking code:
https://gist.github.com/hochgi/596050e7249d2dd0118d7c0a77e1056b

(calculating Collatz conjecture (3n+1 problem) on the first 1000 numbers.)
this shows that the GraphStage approach has significantly better results.

sample run on my laptop:
graphTime: 27.04429695 ms
graphStageTime: 17.330461899999996 ms

Since I plan to use this stage in another PR (first draft in gist: https://gist.github.com/hochgi/c8ee3273034fff8958d306459072fb22#file-akka-stream-retry-md), 
which at least for me, is very important to be as efficient as possible,
I'm still leaning towards the GraphStage approach.

So, any chance you'll take the PR as is?

src/main/scala/akka/stream/contrib/PartitionWith.scala
```diff
@@ -0,0 +1,92 @@
...
+ * @tparam Out1 right output type
+ */
+final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1]) extends GraphStage[FanOutShape2[In, Out0, Out1]] {
+
+  override val shape = new FanOutShape2[In, Out0, Out1]("partitionWith")
+
+  override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+    import shape._
+
+    private var pending = null.asInstanceOf[Either[Out0, Out1]]
```
patriknw: it would be more clean with 
```
private var pending: [Either[Out0, Out1]] = null
```

then you can also use `pending = null` below at two places

patriknw:
LGTM

drewhk:
Ok, I am fine with the stage, although we should prefer combinations instead.

hochgi:
@pariknw @drewhk great. I'll ammend the cleaner code later today (currently I'm AFK). Thanks :)

hochgi:
@drewhk @patriknw @ktoso Iv'e ammended the requested changes, and forced push it.
Not sure how to see diffs of ammended commits in github, but here's what the command line outputs:

```diff
diff --git a/src/main/scala/akka/stream/contrib/PartitionWith.scala b/src/main/scala/akka/stream/contrib/PartitionWith.scala
index 3c51ff7..ccae9ba 100644
--- a/src/main/scala/akka/stream/contrib/PartitionWith.scala
+++ b/src/main/scala/akka/stream/contrib/PartitionWith.scala
@@ -52,7 +52,10 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
   override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
     import shape._
 
-    private var pending = null.asInstanceOf[Either[Out0, Out1]]
+    private var pending: Either[Out0, Out1] = null
+
+    @inline private def downstreamDemand: Boolean =
+      isAvailable(out0) || isAvailable(out1)
 
     setHandler(in, new InHandler {
       override def onPush() = {
@@ -62,7 +65,8 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
           case Right(o) if isAvailable(out1) => push(out1, o)
           case either                        => pending = either
         }
-        if (pending ne null) pull(in)
+        if (pending.eq(null) && downstreamDemand)
+          pull(in)
       }
 
       override def onUpstreamFinish() = {
@@ -75,7 +79,7 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
       override def onPull() = if (pending ne null) pending.left.foreach { o =>
         push(out0, o)
         if (isClosed(in)) completeStage()
-        else pending = null.asInstanceOf[Either[Out0, Out1]]
+        else pending = null
       }
       else if (!hasBeenPulled(in)) pull(in)
     })
@@ -84,7 +88,7 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
       override def onPull() = if (pending ne null) pending.right.foreach { o =>
         push(out1, o)
         if (isClosed(in)) completeStage()
-        else pending = null.asInstanceOf[Either[Out0, Out1]]
+        else pending = null
       }
       else if (!hasBeenPulled(in)) pull(in)
     })
```

src/main/scala/akka/stream/contrib/PartitionWith.scala
```diff
@@ -0,0 +1,96 @@
...
+        if (pending eq null)
+          completeStage()
+      }
+    })
+
+    setHandler(out0, new OutHandler {
+      override def onPull() = if (pending ne null) pending.left.foreach { o =>
+        push(out0, o)
+        if (isClosed(in)) completeStage()
+        else pending = null
```
hochgi: Shouldn't we check `isAvailable(out1)` inside else, and if so, we should `pull(in)`?

src/main/scala/akka/stream/contrib/PartitionWith.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+      else if (!hasBeenPulled(in)) pull(in)
+    })
+
+    setHandler(out1, new OutHandler {
+      override def onPull() = if (pending ne null) pending.right.foreach { o =>
+        push(out1, o)
+        if (isClosed(in)) completeStage()
+        else pending = null
```
hochgi: Same as comment in line 82, but for out0

hochgi:
fixed a a little bug. diff:
```diff
diff --git a/src/main/scala/akka/stream/contrib/PartitionWith.scala b/src/main/scala/akka/stream/contrib/PartitionWith.scala
index ccae9ba..c7d281b 100644
--- a/src/main/scala/akka/stream/contrib/PartitionWith.scala
+++ b/src/main/scala/akka/stream/contrib/PartitionWith.scala
@@ -54,19 +54,21 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
 
     private var pending: Either[Out0, Out1] = null
 
-    @inline private def downstreamDemand: Boolean =
-      isAvailable(out0) || isAvailable(out1)
-
     setHandler(in, new InHandler {
       override def onPush() = {
         val elem = grab(in)
         p(elem) match {
-          case Left(o) if isAvailable(out0)  => push(out0, o)
-          case Right(o) if isAvailable(out1) => push(out1, o)
-          case either                        => pending = either
+          case Left(o) if isAvailable(out0) =>
+            push(out0, o)
+            if (isAvailable(out1))
+              pull(in)
+          case Right(o) if isAvailable(out1) =>
+            push(out1, o)
+            if (isAvailable(out0))
+              pull(in)
+          case either =>
+            pending = either
         }
-        if (pending.eq(null) && downstreamDemand)
-          pull(in)
       }
 
       override def onUpstreamFinish() = {
@@ -79,7 +81,11 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
       override def onPull() = if (pending ne null) pending.left.foreach { o =>
         push(out0, o)
         if (isClosed(in)) completeStage()
-        else pending = null
+        else {
+          pending = null
+          if (isAvailable(out1))
+            pull(in)
+        }
       }
       else if (!hasBeenPulled(in)) pull(in)
     })
@@ -88,7 +94,11 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
       override def onPull() = if (pending ne null) pending.right.foreach { o =>
         push(out1, o)
         if (isClosed(in)) completeStage()
-        else pending = null
+        else {
+          pending = null
+          if (isAvailable(out0))
+            pull(in)
+        }
       }
       else if (!hasBeenPulled(in)) pull(in)
     })
```

I think it's now ready to be merged.

patriknw:
looks like this is ready, I'm merging

patriknw:
Thanks for contributing, @hochgi 
More is welcome.
I'll see if there is anything more to be merged here before releasing, otherwise I'll cut a new release.

PR #23, opened 2016-06-14
WIP: add sample stage

zhxiaogg:
ref #21 

I started by a simple implementation. 

questions:
- I tried `def apply(maxStep:Int, random:Random=Random)` for random sampling, but looks wired. Any suggestion?
- need to implement time-based sampling?

Feel free to review or close it.

### DIFF
--- a/src/main/scala/akka/stream/contrib/Sample.scala
+++ b/src/main/scala/akka/stream/contrib/Sample.scala
@@ -1,21 +1,35 @@
 package akka.stream.contrib
 
+import java.util.Random
+import java.util.concurrent.ThreadLocalRandom
+
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 
-object Sample
-{
+
+object Sample {
   /**
     *
     * returns every nth elements
     *
-    * @param nth
+    * @param nth must > 0
     * @tparam T
     * @return
     */
   def apply[T](nth: Int): Sample[T] = Sample[T](() => nth)
 
-  // def random[T](maxStep: Int = 1000, random: Random = Random): Sample[T] = Sample[T](() => random.nextInt(maxStep))
+  /**
+    *
+    * randomly sampling on a stream
+    *
+    * @param maxStep must > 0, default 1000, the randomly step will be between 1 (inclusive) and maxStep (inclusive)
+    * @tparam T
+    * @return
+    */
+  def random[T](maxStep: Int = 1000): Sample[T] = {
+    require(maxStep > 0, "max step for a random sampling must > 0")
+    Sample[T](() => ThreadLocalRandom.current().nextInt(maxStep) + 1)
+  }
 }
 
 
@@ -25,40 +39,36 @@ object Sample
   * @param next a lambda returns next sample position
   * @tparam T
   */
-case class Sample[T](next: () => Int) extends GraphStage[FlowShape[T, T]]
-{
+case class Sample[T](next: () => Int) extends GraphStage[FlowShape[T, T]] {
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler {
-    var step    = next()
+    var step = getNextStep()
     var counter = 0
 
-    @throws[Exception](classOf[Exception])
-    override def onPull(): Unit = {
-      if (!hasBeenPulled(in)) {
-        pull(in)
-      }
+    def onPull(): Unit = {
+      pull(in)
     }
 
-    @throws[Exception](classOf[Exception])
-    override def onUpstreamFinish(): Unit = {
-      completeStage()
-    }
-
-    @throws[Exception](classOf[Exception])
-    override def onPush(): Unit = {
+    def onPush(): Unit = {
       counter += 1
-      if (counter == step) {
+      if (counter >= step) {
         counter = 0
-        step = next()
+        step = getNextStep()
         push(out, grab(in))
       } else {
         pull(in)
       }
     }
 
+    private def getNextStep(): Long = {
+      val nextStep = next()
+      require(nextStep > 0, s"sampling step should be a positive value: ${nextStep}")
+      nextStep
+    }
+
     setHandlers(in, out, this)
   }
 
-  private  val in    = Inlet[T]("Sample-in")
-  private  val out   = Outlet[T]("Sample-out")
+  val in = Inlet[T]("Sample-in")
+  val out = Outlet[T]("Sample-out")
   override val shape = FlowShape(in, out)
 }
--- a/src/main/scala/akka/stream/contrib/Sample.scala
+++ b/src/main/scala/akka/stream/contrib/Sample.scala
@@ -1,0 +1,64 @@
+package akka.stream.contrib
+
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+
+object Sample
+{
+  /**
+    *
+    * returns every nth elements
+    *
+    * @param nth
+    * @tparam T
+    * @return
+    */
+  def apply[T](nth: Int): Sample[T] = Sample[T](() => nth)
+
+  // def random[T](maxStep: Int = 1000, random: Random = Random): Sample[T] = Sample[T](() => random.nextInt(maxStep))
+}
+
+
+/**
+  * supports sampling on stream
+  *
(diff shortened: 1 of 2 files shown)


src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+        pull(in)
+      }
+    }
+
+    @throws[Exception](classOf[Exception])
+    override def onUpstreamFinish(): Unit = {
+      completeStage()
+    }
+
+    @throws[Exception](classOf[Exception])
```
patriknw: these throws should not be needed, since this is scala

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+    override def onPush(): Unit = {
+      counter += 1
+      if (counter == step) {
+        counter = 0
+        step = next()
```
patriknw: I don't think this will work when `Int` wraps around. Easiest would be to change the `next` function to `Long`

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+  */
+case class Sample[T](next: () => Int) extends GraphStage[FlowShape[T, T]]
+{
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler {
+    var step    = next()
+    var counter = 0
+
+    @throws[Exception](classOf[Exception])
+    override def onPull(): Unit = {
+      if (!hasBeenPulled(in)) {
```
patriknw: this check is not needed

src/test/scala/akka/stream/contrib/SampleSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+  "Sample Stage" should {
+    "returns every Nth element in stream" in {
+      val list = 1 to 1000
+      val source = Source.fromIterator[Int](() => list.toIterator)
+
+      for (n <- 1 to 100) {
+        val future = source.via(Sample(n)).runFold(List.empty[Int])((list, e) => e :: list)
+        val expected = list.filter(_ % n == 0).reverse.toList
+
+        Await.result(future, 200 millis) should ===(expected)
```
patriknw: Don't use that short timeouts if you don't have a good reason. CI on Travis can be very slow. 3 seconds is a good default.

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+    *
+    * returns every nth elements
+    *
+    * @param nth
+    * @tparam T
+    * @return
+    */
+  def apply[T](nth: Int): Sample[T] = Sample[T](() => nth)
+
+  // def random[T](maxStep: Int = 1000, random: Random = Random): Sample[T] = Sample[T](() => random.nextInt(maxStep))
```
patriknw: instead of `scala.Random` I think it should be `java.util.Random`.

I think the default implementation should use `java.util.concurrent.ThreadLocalRandom`, i.e.

    def random[T](maxStep: Int = 1000): Sample[T] = Sample[T](() => ThreadLocalRandom.current().nextInt(maxStep))

patriknw: it's good that you don't use several `apply` overloads

patriknw:
looking good, @zhxiaogg

src/main/scala/akka/stream/contrib/Sample.scala
drewhk: Alternative is to take a function `() => Int`, since that does not tie the random sampling to a particular interface. Then the default can point to a cached function using a random generator (possibly `ThreadLocalRandom`, although these are very low quality random numbers).

src/main/scala/akka/stream/contrib/Sample.scala
drewhk: yep, remove it.

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+    @throws[Exception](classOf[Exception])
+    override def onPull(): Unit = {
+      if (!hasBeenPulled(in)) {
+        pull(in)
+      }
+    }
+
+    @throws[Exception](classOf[Exception])
+    override def onUpstreamFinish(): Unit = {
+      completeStage()
```
drewhk: this is the default anyway, no need for the override.

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,64 @@
...
+        push(out, grab(in))
+      } else {
+        pull(in)
+      }
+    }
+
+    setHandlers(in, out, this)
+  }
+
+  private  val in    = Inlet[T]("Sample-in")
```
drewhk: make these public. That allows using the stage in a graphDSL block directly.

src/test/scala/akka/stream/contrib/SampleSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+  private implicit val system       = ActorSystem("SampleTest")
+  private implicit val materializer = ActorMaterializer()
+
+  "Sample Stage" should {
+    "returns every Nth element in stream" in {
+      val list = 1 to 1000
+      val source = Source.fromIterator[Int](() => list.toIterator)
+
+      for (n <- 1 to 100) {
+        val future = source.via(Sample(n)).runFold(List.empty[Int])((list, e) => e :: list)
```
drewhk: use `Sink.toSeq` instead

src/test/scala/akka/stream/contrib/SampleSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+        val expected = list.filter(_ % n == 0).reverse.toList
+
+        Await.result(future, 200 millis) should ===(expected)
+      }
+    }
+
+    "returns elements randomly" in {
```
drewhk: Add a test case where the returned integer is negative and also when it is zero.

drewhk:
Small improvements needed, otherwise looks good.

johanandren:
LGTM

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,74 @@
...
+    *
+    * randomly sampling on a stream
+    *
+    * @param maxStep must > 0, default 1000, the randomly step will be between 1 (inclusive) and maxStep (inclusive)
+    * @tparam T
+    * @return
+    */
+  def random[T](maxStep: Int = 1000): Sample[T] = {
+    require(maxStep > 0, "max step for a random sampling must > 0")
+    Sample[T](() => ThreadLocalRandom.current().nextInt(maxStep) + 1)
```
drewhk: Unfortunately the numbers returned by this call are super-bad quality from a scientific point of view. I think this should be detailed in the scaladoc and recommend people that need really good behaved sampling to use their own (XorShift family for example) explicitly.

drewhk: Also, this does not provide truly independent sampling either, so maybe call this `randomStep` instead of `random`.

drewhk:
I don't want to prolong this review forever, but I feel an interface `T => Boolean` instead of `T => Int` can be more useful if mutable state is allowed in the function. That can model both the current sampling methods, and also new ones that are not covered by these. WDYT?

2m:
@zhxiaogg can you address latest comments and move these two files under `contrib` directory?

zhxiaogg:
@2m sure, no problem.

@drewhk  I think the current interface of Sample case class is enough for people to inject their own sampling logic (eg. XorShift). 
My first thought is that sampling is based on time or space, irrelevant  of what the element is or looks like. So the current constructor parameter of Sample is actually `() => Int`(based on space), not `T => Int`. Besides if we change it to `T => Boolean`, we can use `filter` directly and eventually there seems no need of Sample.

drewhk:
Ok, thanks for explaining. LGTM

ktoso:
It's useful enough, thanks @zhxiaogg - LGTM

src/main/scala/akka/stream/contrib/Sample.scala
```diff
@@ -0,0 +1,74 @@
+package akka.stream.contrib
```
2m: We do. Missing header here was an indication that sbt did not see that file. And that was because it was not moved to the `contrib` dir. I created a PR to fix it: https://github.com/akka/akka-stream-contrib/pull/55

Issue #25, opened 2016-06-20
Add a retry graph factory, that works on flows


hochgi:
as discussed briefly on gitter, a really helpful API, would be a generic (generic as it can be) retry factory for flows.

1st draft (with some known issues, I haven't fixed yet) is described in this gist:
https://gist.github.com/hochgi/c8ee3273034fff8958d306459072fb22#file-akka-stream-retry-md

also, a proposed API will be submitted as a PR shortly (without implementation just yet).
I would like this issue/PR to be an open discussion on what would be the best API ("best" in terms of simplicity, generic, usefulness, etc'...).

also, if this contribution will get into the main streams code, it could be useful to have retry as a combinator (similar to how `via` is used).

PR #26, opened 2016-06-20
Retry flows API

hochgi:
PR for https://github.com/akka/akka-stream-contrib/issues/25

### DIFF
--- a/src/main/scala/akka/stream/contrib/Retry.scala
+++ b/src/main/scala/akka/stream/contrib/Retry.scala
@@ -1,0 +1,251 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream._
+import akka.stream.scaladsl._
+import akka.stream.stage._
+import scala.util.{ Try, Success, Failure }
+import scala.concurrent.Future
+import scala.collection.immutable
+
+/**
+ * This object defines a factory methods for retry operaions.
+ */
+object Retry {
+
+  /**
+   * Retry flow factory. given a flow that produces `Try`s, this wrapping flow may be used to try
+   * and pass failed elements through the flow again. More accurately, the given flow consumes a tuple
+   * of `input` & `state`, and produces a tuple of `Try` of `output` and `state`.
+   * If the flow emits a failed element (i.e. `Try` is a `Failure`), the `retryWith` function is fed with the
+   * `state` of the failed element, and may produce a new input-state tuple to pass through the original flow.
+   * The function may also yield `None` instead of `Some((input,state))`, which means not to retry a failed element.
+   *
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result,
+   * and may result in a deadlock.
+   *
+   * @param flow the flow to retry
+   * @param retryWith if output was failure, we can optionaly recover from it,
+   *        and retry with a new pair of input & new state we get from this function.
+   * @tparam I input elements type
+   * @tparam O output elements type
+   * @tparam S state to create a new `(I,S)` to retry with
+   * @tparam M materialized value type
+   */
+  def apply[I, O, S, M](flow: Graph[FlowShape[(I, S), (Try[O], S)], M])(retryWith: S => Option[(I, S)]): Graph[FlowShape[(I, S), (Try[O], S)], M] = {
+    GraphDSL.create(flow) { implicit b => origFlow =>
+      import GraphDSL.Implicits._
+
+      val retry = b.add(new RetryCoordinator[I, S, O](retryWith))
+
+      retry.out2 ~> origFlow ~> retry.in2
+
+      FlowShape(retry.in1, retry.out1)
+    }
+  }
+
+  /**
+   * Factory for multiple retries flow. similar to the simple retry, but this will allow to
+   * break down a "heavy" element which failed into multiple "thin" elements, that may succeed individually.
+   * Since it's easy to inflate elements in retry cycle, there's also a limit parameter,
+   * that will limit the amount of generated elements by the `retryWith` function,
+   * and will fail the stage if that limit is exceeded.
+   * Passing `Some(Nil)` is valid, and will result in filtering out the failure quietly, without
+   * emitting a failed `Try` element.
+   *
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result,
+   * and may result in a deadlock.
+   *
+   * @param limit since every retry can generate more elements,
+   *        the inner queue can get too big. if the limit is reached,
+   *        the stage will fail.
+   * @param flow the flow to retry
+   * @param retryWith if output was failure, we can optionaly recover from it,
+   *        and retry with a sequence of input & new state pairs we get from this function.
+   * @tparam I input elements type
+   * @tparam O output elements type
+   * @tparam S state to create a new `(I,S)` to retry with
+   * @tparam M materialized value type
+   */
+  def concat[I, O, S, M](limit: Long, flow: Graph[FlowShape[(I, S), (Try[O], S)], M])(retryWith: S => Option[immutable.Iterable[(I, S)]]): Graph[FlowShape[(I, S), (Try[O], S)], M] = {
+    GraphDSL.create(flow) { implicit b => origFlow =>
+      import GraphDSL.Implicits._
+
+      val retry = b.add(new RetryConcatCoordinator[I, S, O](limit, retryWith))
+
+      retry.out2 ~> origFlow ~> retry.in2
+
+      FlowShape(retry.in1, retry.out1)
+    }
+  }
+
+  private[akka] class RetryCoordinator[I, S, O](retryWith: S => Option[(I, S)]) extends GraphStage[BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)]] {
+    val in1 = Inlet[(I, S)]("Retry.ext.in")
+    val out1 = Outlet[(Try[O], S)]("Retry.ext.out")
+    val in2 = Inlet[(Try[O], S)]("Retry.int.in")
+    val out2 = Outlet[(I, S)]("Retry.int.out")
+    override val shape = BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)](in1, out1, in2, out2)
+    override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+      var elementInCycle = false
+      var pending: (I, S) = null
+
+      setHandler(in1, new InHandler {
+        override def onPush() = {
+          val is = grab(in1)
+          if (!hasBeenPulled(in2)) pull(in2)
+          push(out2, is)
+          elementInCycle = true
+        }
+
+        override def onUpstreamFinish() = {
+          if (!elementInCycle)
+            completeStage()
+        }
+      })
+
+      setHandler(out1, new OutHandler {
+        override def onPull() = {
+          if (isAvailable(out2)) pull(in1)
+          else pull(in2)
+        }
+      })
+
+      setHandler(in2, new InHandler {
+        override def onPush() = {
+          elementInCycle = false
+          grab(in2) match {
+            case s @ (_: Success[O], _) => pushAndCompleteIfLast(s)
+            case failure @ (_, s) => retryWith(s).fold(pushAndCompleteIfLast(failure)) { is =>
+              pull(in2)
+              if (isAvailable(out2)) {
+                push(out2, is)
+                elementInCycle = true
+              } else pending = is
+            }
+          }
+        }
+      })
+
+      def pushAndCompleteIfLast(elem: (Try[O], S)): Unit = {
+        push(out1, elem)
+        if (isClosed(in1))
+          completeStage()
+      }
+
+      setHandler(out2, new OutHandler {
+        override def onPull() = {
+          if (isAvailable(out1) && !elementInCycle) {
+            if (pending ne null) {
+              push(out2, pending)
+              pending = null
+              elementInCycle = true
+            } else if (!hasBeenPulled(in1)) {
+              pull(in1)
+            }
+          }
+        }
+
+        override def onDownstreamFinish() = {
+          //Do Nothing, intercept completion as downstream
+        }
+      })
+    }
+  }
+
+  private[akka] class RetryConcatCoordinator[I, S, O](limit: Long, retryWith: S => Option[immutable.Iterable[(I, S)]]) extends GraphStage[BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)]] {
+    val in1 = Inlet[(I, S)]("RetryConcat.ext.in")
+    val out1 = Outlet[(Try[O], S)]("RetryConcat.ext.out")
+    val in2 = Inlet[(Try[O], S)]("RetryConcat.int.in")
+    val out2 = Outlet[(I, S)]("RetryConcat.int.out")
+    override val shape = BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)](in1, out1, in2, out2)
+    override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+      var elementInCycle = false
+      val queue = scala.collection.mutable.Queue.empty[(I, S)]
+
+      setHandler(in1, new InHandler {
+        override def onPush() = {
+          val is = grab(in1)
+          if (!hasBeenPulled(in2)) pull(in2)
+          if (isAvailable(out2)) {
+            push(out2, is)
+            elementInCycle = true
+          } else queue.enqueue(is)
+        }
+
+        override def onUpstreamFinish() = {
+          if (!elementInCycle && queue.isEmpty)
+            completeStage()
+        }
+      })
+
+      setHandler(out1, new OutHandler {
+        override def onPull() = {
+          if (queue.isEmpty) {
+            if (isAvailable(out2)) pull(in1)
+            else pull(in2)
+          } else {
+            pull(in2)
+            if (isAvailable(out2)) {
+              push(out2, queue.dequeue())
+              elementInCycle = true
+            }
+          }
+        }
+      })
+
+      setHandler(in2, new InHandler {
+        override def onPush() = {
+          elementInCycle = false
+          grab(in2) match {
+            case s @ (_: Success[O], _) => pushAndCompleteIfLast(s)
+            case failure @ (_, s) => retryWith(s).fold(pushAndCompleteIfLast(failure)) { xs =>
+              if (xs.size + queue.size > limit) failStage(new IllegalStateException(s"Queue limit of $limit has been exceeded. Trying to append ${xs.size} elements to a queue that has ${queue.size} elements."))
+              else {
+                xs.foreach(queue.enqueue(_))
+                if (queue.isEmpty) {
+                  if (isClosed(in1)) completeStage()
+                  else pull(in1)
+                } else {
+                  pull(in2)
+                  if (isAvailable(out2)) {
+                    push(out2, queue.dequeue())
+                    elementInCycle = true
+                  }
+                }
+              }
+            }
+          }
+        }
+      })
+
+      def pushAndCompleteIfLast(elem: (Try[O], S)): Unit = {
+        push(out1, elem)
+        if (isClosed(in1) && queue.isEmpty)
+          completeStage()
+      }
+
+      setHandler(out2, new OutHandler {
+        override def onPull() = {
+          if (!elementInCycle && isAvailable(out1)) {
+            if (queue.isEmpty) pull(in1)
+            else {
+              push(out2, queue.dequeue())
+              elementInCycle = true
+              if (!hasBeenPulled(in2)) pull(in2)
+            }
+          }
+        }
+
+        override def onDownstreamFinish() = {
+          //Do Nothing, intercept completion as downstream
+        }
+      })
+    }
+  }
+}
--- a/src/test/scala/akka/stream/contrib/RetrySpec.scala
+++ b/src/test/scala/akka/stream/contrib/RetrySpec.scala
@@ -1,0 +1,362 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.KillSwitches
+import akka.stream.scaladsl._
+import akka.stream.testkit.scaladsl.{ TestSink, TestSource }
+
+import scala.concurrent.duration._
+import scala.util.{ Failure, Success, Try }
+
+class RetrySpecAutoFusingOn extends { val autoFusing = true } with RetrySpec
+class RetrySpecAutoFusingOff extends { val autoFusing = false } with RetrySpec
+
+trait RetrySpec extends BaseStreamSpec {
+
+  val failedElem: Try[Int] = Failure(new Exception("cooked failure"))
+  def flow[T] = Flow[(Int, T)].map {
+    case (i, j) if i % 2 == 0 => (failedElem, j)
+    case (i, j)               => (Success(i + 1), j)
+  }
+
+  "Retry" should {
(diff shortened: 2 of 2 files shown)


hochgi:
Is the `concat` APIs possible? If we have the retries cycling in a loop, adding more elements in a closed loop would require unbounded buffering or it would deadlock.

Maybe it can be done, but not with the suggested API. We must not add elements to the loop unless the same amount of elements is removed from the loop.

So just playing around, what if we add a special type:

```scala
sealed trait RetryFork[T]
case object Emit extends RetryFork[Nothing]
case class Retry[T](head: T, tail: RetryFork[T]) extends RetryFork[T]
case class EmitAndRetry[T](head: T, tail: RetryFork[T]) extends RetryFork[T]
```

And have the state `S` be of this type, resulting in an API similar to:

```scala
def concat[I,O,S,M](flow: Flow[(I,S),(Try[O],S),M])(retryWith: S => RetryFork[(I,S)])
```
 Thus enabling to generate more elements to drop into the retry loop, but doing so lazily, so new elements are only generated when a previous element left the cycle

Not sure all of this makes sense, and probably an overkill for a generic API, but just thinking out loud here...

viktorklang:
There is no generic way to know if an element leaves the loop.

On Wed, Jun 22, 2016 at 3:51 AM, Gilad Hoch <notifications@github.com>
wrote:

> Is the concat APIs possible? If we have the retries cycling in a loop,
> adding more elements in a closed loop would require unbounded buffering or
> it would deadlock.
>
> Maybe it can be done, but not with the suggested API. We must not add
> elements to the loop unless the same amount of elements is removed from the
> loop.
>
> So just playing around, what if we add a special type:
>
> sealed trait RetryFork[T]case object Emit extends RetryFork[Nothing]case class Retry[T](head: T, tail: RetryFork[T]) extends RetryFork[T]case class EmitAndRetry[T](head: T, tail: RetryFork[T]) extends RetryFork[T]
>
> And have the state S be of this type, resulting in an API similar to:
>
> def concat[I,O,S,M](flow: Flow[(I,S),(Try[O],S),M])(retryWith: S => RetryFork[(I,S)])
>
> Thus enabling to generate more elements to drop into the retry loop, but
> doing so lazily, so new elements are only generated when a previous element
> left the cycle
>
> Not sure all of this makes sense, and probably an overkill for a generic
> API, but just thinking out loud here...
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/akka/akka-stream-contrib/pull/26#issuecomment-227622133>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAqd1jyP6JvtiT-rKZEbooMTNpR-8BTks5qOJU_gaJpZM4I5cEY>
> .
>



-- 
Cheers,
√

hochgi:
@viktorklang not for a specific element, but we can wrap the entire graph wih stages that keep track of how many element are in the cycle. e.g:
retry graph

viktorklang:
@hochgi No, I meant, for a Graph X, there is no generic way to track element "loss" :)

hochgi:
you mean inside the given flow? a flow isn't limited to 1:1 input-output, but that will be of course documented as a caveat users should be aware to.

What I have in mind, is the obvious usecase (that got me into this in the first place) of a connection pool flow.
something like `Http().superPool()`, and I assume this flow is guaranteed to be 1:1 input-output.

viktorklang:
@hochgi Yes, the reason I say this is because if all a user has is a Flow[I,O,M] it is impossible to know if it is 1:1 or not.

hochgi:
@viktorklang OK, I'm fine with documenting this as a warning. it's something the user can, but should not do. like passing `null`s...

viktorklang:
@hochgi passing nulls as elements is not allowed :)

hochgi:
current status: tried implementing the basic functionality described in the image
retry graph
but stream won't shut down. implementation:
https://github.com/hochgi/akka-stream-contrib/commit/9812757ff1103eaa4335a4c10212e2ad531672a1
not sure what I missed, will dig into it later.

hochgi:
The simple retry works now. currently thinking how to implement the `retryConcat`, and if the API is reasonable.

hochgi:
in the last amended commit I removed the async APIs, and implemented the concat according to this diagram:
retry concat graph
there is already test and some very minimal documentation.
To get this PR ready for merge I want to add more docs.
Especially, I want to document the caveats discussed here (e.g: Must provide 1:1 flows)
other than that, I think the code is ready for reviews.
cc @viktorklang

hochgi:
Documentation is written, and commits are squashed.

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,245 @@
...
+
+      override def preStart() = pull(in1)
+
+      setHandler(in0, new InHandler {
+        override def onPush() = {
+          push(out, grab(in0))
+          elementsInCycleCount += 1
+        }
+
+        override def onUpstreamFinish() = {
```
patriknw: is there a need for `setKeepGoing(true)`?

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,245 @@
...
+      FlowShape(elemQueue.in0, coElemQueue.out0)
+    }
+  }
+
+  private[akka] class CountElementsAndInterceptCompletion[T] extends GraphStage[FanInShape2[T, Int, T]] {
+    override val shape = new FanInShape2[T, Int, T]("countElementsAndInterceptCompletion")
+    override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+
+      import shape._
+      var elementsInCycleCount = 0
```
patriknw: `Long` to avoid (theoretical) `Int` wrap around issues?

src/main/scala/akka/stream/contrib/Retry.scala
hochgi: I don't think so. in case there are no elements in the cycle, the stage should complete regularly.

src/main/scala/akka/stream/contrib/Retry.scala
hochgi: OK, it's not likely, but is possible that the given flow will aggregate more than `Int.MaxValue` elements. I will change to `Long`

hochgi:
@patriknw changed `Int` to `Long` and amended the commit :)

drewhk:
You can wrap the Flow with a BidiFlow that tracks that not more elements are emitted than have been consumed, otherwise fails the stream.

drewhk:
nice diagrams btw.

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,245 @@
...
+import scala.collection.immutable
+
+/**
+ * This object defines a factory methods for retry operaions.
+ */
+object Retry {
+
+  /**
+   * Factory for a retry flow. Given a flow which recieves a tuple of "input" & "state",
+   * and emits a tuples of a Try of "output" & "state", and given a function that given a state
```
drewhk: very, very long sentence :) Please try to refactor.

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,245 @@
...
+      elemQueue.in1       <~       partition.out1
+      elemQueue.in2                  <~              coElemQueue.out1
+      // format: ON
+      FlowShape(elemQueue.in0, coElemQueue.out0)
+    }
+  }
+
+  private[akka] class CountElementsAndInterceptCompletion[T] extends GraphStage[FanInShape2[T, Long, T]] {
+    override val shape = new FanInShape2[T, Long, T]("countElementsAndInterceptCompletion")
+    override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
```
drewhk: extend OutHandler to reduce allocations

drewhk:
One option for improvement is to make the increment/decrement stages one BidiStage instead, then there is no cost of communication (and possible races) between the two sides.

src/test/scala/akka/stream/contrib/RetrySpec.scala
```diff
@@ -0,0 +1,71 @@
...
+trait RetrySpec extends BaseStreamSpec {
+
+  val failedElem: Try[Int] = Failure(new Exception("cooked failure"))
+  val flow = Flow[(Int, Int)].map {
+    case (i, j) if i % 2 == 0 => (failedElem, j)
+    case (i, j)               => (Success(i + 1), j)
+  }
+
+  "Retry" should {
+    "retry ints according to their parity" in {
```
drewhk: more tests are needed, especially around completion, and also testing deadlock-freedom.

drewhk:
I am not sure that this graph is necessary. I think it might be far simpler to concentrate all the logic and bookkeeping to one BidiStage. It would look like this:

```
       +-------------------+
 +-----> Flow              +----+
 |     +-------------------+    |
 |                              |
 |     +-------------------+    |
 +-----+ RertryCoordinator +----+
       |                   |
 +----->                   +---->
       +-------------------+
```

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,245 @@
...
+      FlowShape(elemCount.in0, coElemCount.out0)
+    }
+  }
+
+  /**
+   * Factory for multiple retries flow. similar to the simple retry, but this will allow to
+   * break down a "heavy" element which failed into multiple "thin" elements, that may succeed
+   * indevidually. Since it's very easy to inflate elements in retry cycle, there's also a
+   * limit parameter, that will limit the amount of generated elements by the `retryWith` function,
+   * and will fail the stage if that limit is exceeded.
```
hochgi: Maybe the API can be expanded to receive an overflow strategy, like the buffer stage.
I have thought of that, and decided it's an overkill for now.  But since it's the contrib, I assume that if down the road we break API and add another overflow strategy parameter, it'll be OK.

anyway, would you want me to add that now before accepting this PR?

P.S. as a default behaviour, IMO, failing the stage is better. as a user of this library, I would want to know in the development phase I did something wrong, instead of discovering in production I'm losing data silently.

src/main/scala/akka/stream/contrib/Retry.scala
hochgi: Not sure what you mean... extends where?

hochgi:
@drewhk the BidiFlow idea is great! wish I'd thought of it earlier...
regarding failing the stage if the given flow output more elements than it should, I think it's not a good idea.
you can only detect if the are too many elements coming out, but what about a flow that reduces the number of elements? If we can't guarantee this safely, as @viktorklang pointed out, I think we shouldn't do it at all, and leave this as the user responsibility.

src/test/scala/akka/stream/contrib/RetrySpec.scala
hochgi: indeed. will do

src/main/scala/akka/stream/contrib/Retry.scala
drewhk: I meant `new GraphStageLogic(shape) with OutHandler` and then `setHandler(out, this)`.

src/main/scala/akka/stream/contrib/Retry.scala
drewhk: ok, agreed!

drewhk:
> regarding failing the stage if the given flow output more elements than it should, I think it's not a good idea.
you can only detect if the are too many elements coming out, but what about a flow that reduces the number of elements? If we can't guarantee this safely, as @viktorklang pointed out, I think we shouldn't do it at all, and leave this as the user responsibility.

Ok, you convinced me.

hochgi:
@drewhk I have implemented from scratch using a bidi flow. that was a great idea! thanks :)
So now, I just need to write some additional tests & improve documentation.

drewhk:
Ok, now I look at this finally :)

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,244 @@
...
+      val retry = b.add(new RetryConcatCoordinator[I, S, O](limit, retryWith))
+
+      retry.out2 ~> origFlow ~> retry.in2
+
+      FlowShape(retry.in1, retry.out1)
+    }
+  }
+
+  private[akka] class RetryCoordinator[I, S, O](retryWith: S => Option[(I, S)]) extends GraphStage[BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)]] {
+    val in1 = Inlet[(I, S)]("RertryCoordinator.in")
```
drewhk: I recommend using more descriptive names for these ports to so the code reads nicer.

hochgi: Any suggestions?

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,244 @@
...
+
+        override def onUpstreamFinish() = {
+          if (!elementInCycle)
+            completeStage()
+        }
+      })
+
+      setHandler(out1, new OutHandler {
+        override def onPull() =
+          pull(in1)
```
drewhk: it would be good to write up the invariants in a comment. Like the above line relies on the invariant that if `out1` is pulled then `in1` is available for pulling.

drewhk: also, it would be probably helpful to add a comment describing a duty-cycle of the stage.

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,244 @@
...
+    val out2 = Outlet[(I, S)]("RertryCoordinator.cout")
+    override val shape = BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)](in1, out1, in2, out2)
+    override def createLogic(attributes: Attributes) = new GraphStageLogic(shape) {
+      var elementInCycle = false
+      var pending: (I, S) = null
+
+      setHandler(in1, new InHandler {
+        override def onPush() = {
+          val is = grab(in1)
+          if (!hasBeenPulled(in2)) pull(in2)
```
drewhk: hm, does not make sense instead to pull through the wrapped stage? I.e. pulling of out1 results this stage pulling the wrapped flow, which in turn will pull out2, which in turn will pull in1? That is a control flow that might be more understandable.

drewhk: to describe what I mean, the duty cycle with one retry would look like this
 - out1 is pulled by external world, we pull in2 (the wrapped flow)
 - out2 is pulled by wrapped Flow, we pull in1 (the external world)
 - in1 is pushed by external world, we feed result in out2 (to wrapped flow)
 - wrapped flow pushes failure through in2. We decide to retry (instead of pushing to external world), and pull in2 (the wrapped flow)
 - out2 is pulled by wrapped Flow, we provide retry (instead of pulling external world) through out2
 - wrapped flow pushes success through in2, we push it through out1 to external world
 - cycle starts from step 1 again

I quite possibly miss some details here that you already found out, so correct me on any point here.

hochgi: That was my initial logic. But the wrapped flow may pull eagerly without being pulled, always storing one element ready for downstream immediately.

drewhk: Aargh, yes the flow can pre-pull, I forgot that :)

hochgi: well, I think I managed to pull off with your suggested duty cycle (at least on the simpler case, and check if the flow pre-pulls elements)

drewhk: hm, one difference is that on out1.onPull you immediately pull in1, and not the internal Flow.

hochgi: not exactly. in the revised code we have:
```scala
setHandler(out1, new OutHandler {
  override def onPull() = {
    if (isAvailable(out2)) pull(in1)
    else pull(in2)
  }
})
```

So we only pull in1 if the internal flow has already pulled, which means I won't get out2.onPull.
in case out2 isn't available, we do pull in2 and not in2.

drewhk: Oh, why do I don't see the right version here...

drewhk: now I do, weird.

hochgi: yeah... github is being weird lately. I needed to refresh the page a couple of times before getting the new code I committed yesterday :/

src/test/scala/akka/stream/contrib/RetrySpec.scala
```diff
@@ -0,0 +1,264 @@
...
+        .viaMat(KillSwitches.single[Int])(Keep.both)
+        .map(i => (i, i))
+        .via(Retry.concat(100, flow[Int]) { x => Some(List((x, x + 1))) })
+        .toMat(TestSink.probe)(Keep.both)
+        .run()
+
+      killSwitch.abort(failedElem.failed.get)
+      sink.request(1)
+      sink.expectError(failedElem.failed.get)
+    }
```
drewhk: I think it would be important to have a version where the flow passed to Retry itself is terminated via a KillSwitch

drewhk:
The logic looks correct to me, so if a few tests are added where the inner flow is killed I would be happy to merge this :)

hochgi:
There are already 6 such tests. 3 for retry and same 3 for retry.concat
What other scenarios are missing?

src/test/scala/akka/stream/contrib/RetrySpec.scala
hochgi: done :)

drewhk: ok, I am blind. I was looking for something like `Retry(innerFlow.via(KillSwitch))` but `innerFlow` already had it applied, sorry for the noise.

drewhk:
Nothing is missing, I was looking at the wrong place (this time not GHs fault). LGTM

src/test/scala/akka/stream/contrib/RetrySpec.scala
hochgi: No you arent blind. These are new tests :-)

drewhk: ok great, doesn't matter in the end :D

drewhk:
Can you squash the commits into one? Then I would like to merge this (unless other volunteers decide to review).

hochgi:
@drewhk commits are squashed!

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+
+      FlowShape(retry.in1, retry.out1)
+    }
+  }
+
+  private[akka] class RetryCoordinator[I, S, O](retryWith: S => Option[(I, S)]) extends GraphStage[BidiShape[(I, S), (Try[O], S), (Try[O], S), (I, S)]] {
+    val in1 = Inlet[(I, S)]("Rertry.ext.in")
+    val out1 = Outlet[(Try[O], S)]("Rertry.ext.out")
+    val in2 = Inlet[(Try[O], S)]("Rertry.int.in")
+    val out2 = Outlet[(I, S)]("Rertry.int.out")
```
johanandren: Rertry -> Retry in all these

hochgi: @johanandren oops... typo. fixed now :-)

johanandren:
LGTM

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+   * Retry flow factory. given a flow that produces `Try`s, this wrapping flow may be used to try
+   * and pass failed elements through the flow again. More accurately, the given flow consumes a tuple
+   * of "input" & "state", and produces a tuple of `Try` of "output" and "state.
+   * If the flow emits a failed element (i.e. `Try` is a `Failure`), the `retryWith` function is fed with the
+   * "state" of the failed element, and may produce a new "input-state" tuple to pass through the original flow.
+   * The function may also yield `None` instead of `Some((input,state))`, which means not to retry a failed element.
+   *
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result.
```
ktoso: And may result in a deadlock

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * This object defines a factory methods for retry operaions.
+ */
+object Retry {
+
+  /**
+   * Retry flow factory. given a flow that produces `Try`s, this wrapping flow may be used to try
+   * and pass failed elements through the flow again. More accurately, the given flow consumes a tuple
+   * of "input" & "state", and produces a tuple of `Try` of "output" and "state.
```
ktoso:
```
`input`

`state`
```

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result.
+   *
+   * @param flow the flow to retry
+   * @param retryWith if output was failure, we can optionaly recover from it,
+   *        and retry with a new pair of input & new state we get from this function.
+   * @tparam I input elements type
+   * @tparam O output elements type
+   * @tparam S state to create a new (input,state): (I,S) to retry with
+   * @tparam M materialized value type
+   * @return `Graph[FlowShape[(I,S),Try[O]],M]` instance
```
ktoso: remove `@return` doc, it's not helping – we have return types for that ;)

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result.
+   *
+   * @param flow the flow to retry
+   * @param retryWith if output was failure, we can optionaly recover from it,
+   *        and retry with a new pair of input & new state we get from this function.
+   * @tparam I input elements type
+   * @tparam O output elements type
+   * @tparam S state to create a new (input,state): (I,S) to retry with
```
ktoso:
```
@tparam S state to create a new `(I, S)` to retry with
`

File: src/main/scala/akka/stream/contrib/Retry.scala [hunk 16]
```diff
@@ -0,0 +1,251 @@
...
+
+      retry.out2 ~> origFlow ~> retry.in2
+
+      FlowShape(retry.in1, retry.out1)
+    }
+  }
+
+  /**
+   * Factory for multiple retries flow. similar to the simple retry, but this will allow to
+   * break down a "heavy" element which failed into multiple "thin" elements, that may succeed indevidually.
```
ktoso: individually

src/main/scala/akka/stream/contrib/Retry.scala
```diff
@@ -0,0 +1,251 @@
...
+   * Since it's very easy to inflate elements in retry cycle, there's also a limit parameter,
```
ktoso: very easy => easy, we tend to avoid "very" as it depends on who's reading this

ktoso:
LGTM, would be nice to have the graph / docs persited somewhere.
I guess we should think about docs for this project too?

hochgi:
@ktoso I agree. docs for contrib would be great. not sure where it should be though...
streams cookbook?
here in GH as wiki pages?
maybe the new SO documentation would be a good place to start?

ktoso:
> @ktoso I agree. docs for contrib would be great. not sure where it should be though...

No, proper docs – we need docs for Reactive Kafka anyway, so we'll figure out a common way to do it.

> streams cookbook?

No, since it's in the wrong place then

> here in GH as wiki pages?

not wiki pages, docs should be compiled and tested. same as in the akka project.

> maybe the new SO documentation would be a good place to start?

same point as above, reference docs can be compiled and tested + comitted together with initial feature PR. SO is a fun thing, but it's additional.

ktoso:
Anyhow, this feature is a nice one! Seems it was reviewed by 3 already even :-)

Issue #27, opened 2016-07-06
Add an AccumulateWhileUnchanged stage


rucek:
Given a stream of input elements, the stage will observe an arbitrary element property and accumulate the incoming elements while the value of the property remains unchanged. When the property changes, a sequence of accumulated elements will be emitted

PR #28, opened 2016-07-06
Add AccumulateWhileUnchanged

rucek:
As described in #27

### DIFF
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -72,9 +72,7 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
       }
 
       override def onPull(): Unit = {
-        if (!isClosed(in)) {
-          pull(in)
-        }
+        pull(in)
       }
 
       override def onUpstreamFinish(): Unit = {
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -78,9 +78,9 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
       override def onUpstreamFinish(): Unit = {
         val result = buffer.result()
         if (result.nonEmpty) {
-          push(out, result)
+          emit(out, result)
         }
-        super.onUpstreamFinish()
+        completeStage()
       }
     })
 
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -83,5 +83,9 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
         super.onUpstreamFinish()
       }
     })
+
+    override def postStop(): Unit = {
+      buffer.clear()
+    }
   }
 }
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.japi.function
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+
+import scala.collection.immutable
+
+object AccumulateWhileUnchanged {
+
+  /**
+   * Factory for [[AccumulateWhileUnchanged]] instances
+   *
+   * @param propertyExtractor a function to extract the observed element property
+   * @tparam Element  type of accumulated elements
+   * @tparam Property type of the observed property
+   * @return [[AccumulateWhileUnchanged]] instance
+   */
+  def apply[Element, Property](propertyExtractor: Element => Property) = new AccumulateWhileUnchanged(propertyExtractor)
+
+  /**
+   * Java API: Factory for [[AccumulateWhileUnchanged]] instances
+   *
+   * @param propertyExtractor a function to extract the observed element property
+   * @tparam Element  type of accumulated elements
+   * @tparam Property type of the observed property
+   * @return [[AccumulateWhileUnchanged]] instance
+   */
+  def create[Element, Property](propertyExtractor: function.Function[Element, Property]) = new AccumulateWhileUnchanged(propertyExtractor.apply)
+}
+
+/**
+ * Accumulates elements of type [[Element]] while a property extracted with [[propertyExtractor]] remains unchanged,
+ * emits an accumulated sequence when the property changes
+ *
+ * @param propertyExtractor a function to extract the observed element property
+ * @tparam Element  type of accumulated elements
+ * @tparam Property type of the observed property
+ */
+final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Element => Property) extends GraphStage[FlowShape[Element, immutable.Seq[Element]]] {
+
+  val in = Inlet[Element]("AccumulateWhileUnchanged.in")
+  val out = Outlet[immutable.Seq[Element]]("AccumulateWhileUnchanged.out")
+
+  override def shape = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) {
+
+    private var currentState: Option[Property] = None
+    private val buffer = Vector.newBuilder[Element]
+
+    setHandlers(in, out, new InHandler with OutHandler {
+
+      override def onPush(): Unit = {
+        val nextElement = grab(in)
+        val nextState = propertyExtractor(nextElement)
+
+        if (currentState.isEmpty || currentState.contains(nextState)) {
+          buffer += nextElement
+          pull(in)
+        } else {
+          val result = buffer.result()
+          buffer.clear()
+          buffer += nextElement
+          push(out, result)
+        }
+
+        currentState = Some(nextState)
+      }
+
+      override def onPull(): Unit = {
+        if (!isClosed(in)) {
+          pull(in)
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        val result = buffer.result()
+        if (result.nonEmpty) {
+          push(out, result)
+        }
+        super.onUpstreamFinish()
+      }
+    })
+  }
+}
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -58,17 +58,19 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
         val nextElement = grab(in)
         val nextState = propertyExtractor(nextElement)
 
-        if (currentState.isEmpty || currentState.contains(nextState)) {
-          buffer += nextElement
-          pull(in)
-        } else {
-          val result = buffer.result()
-          buffer.clear()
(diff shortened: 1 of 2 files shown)


src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -0,0 +1,89 @@
...
+          buffer.clear()
+          buffer += nextElement
+          push(out, result)
+        }
+
+        currentState = Some(nextState)
+      }
+
+      override def onPull(): Unit = {
+        if (!isClosed(in)) {
```
patriknw: I don't think `in` can be closed here, can it?

src/test/scala/akka/stream/contrib/AccumulateWhileUnchangedSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+  "AccumulateWhileUnchanged" should {
+
+    "emit accumulated elements when the given property changes" in {
+      val (source, sink) = TestSource.probe[Element]
+        .via(AccumulateWhileUnchanged(_.value))
+        .toMat(TestSink.probe)(Keep.both)
+        .run()
+
+      sink.request(42)
+      SampleElements.All.foreach(source.sendNext)
```
patriknw: perhaps use `Source(SampleElements.All)` instead of `TestSource`?

src/test/scala/akka/stream/contrib/AccumulateWhileUnchangedSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      source.sendComplete()
+      sink.expectNext(SampleElements.Ones, SampleElements.Twos, SampleElements.Threes)
+      sink.expectComplete()
+    }
+
+    "not emit any value for an empty source" in {
+      Source(Vector.empty[Element])
```
patriknw: `Source.empty`?

patriknw:
LGTM, useful

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
rucek: Hmm, you're right, I must have had a bug in some intermediate implementation of `onUpstreamFinish`, because at some point it was actually closed and I was getting a "cannot pull from a closed source". Fixing.

src/test/scala/akka/stream/contrib/AccumulateWhileUnchangedSpec.scala
rucek: When I changed to:
```scala
"emit accumulated elements when the given property changes" in {
  val (_, sink) = Source(SampleElements.All)
    .via(AccumulateWhileUnchanged(_.value))
    .toMat(TestSink.probe)(Keep.both)
    .run()

  sink.request(42)
  sink.expectNext(SampleElements.Ones, SampleElements.Twos, SampleElements.Threes)
  sink.expectComplete()
}
```
I'm getting a "cannot push port twice":
```
[info] - should emit accumulated elements when the given property changes *** FAILED *** (13 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected OnNext(List(Element(6,3))), found OnError(java.lang.IllegalArgumentException: requirement failed: Cannot push port (AccumulateWhileUnchanged.out) twice
[info]  at scala.Predef$.require(Predef.scala:224)
[info]  at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:436)
[info]  at akka.stream.contrib.AccumulateWhileUnchanged$$anon$2$$anon$1.onUpstreamFinish(AccumulateWhileUnchanged.scala:81)
```

Any ideas?

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -0,0 +1,89 @@
...
+          pull(in)
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        val result = buffer.result()
+        if (result.nonEmpty) {
+          push(out, result)
```
patriknw: You can't push if you have not been pulled.
You have to set a flag that you are finishing and in `onPull` check that flag and push the final result and then `completeStage`

rucek: I tried with the following:
```scala
override def onPull(): Unit = {
  if (isFinishing) {
    val result = buffer.result()
    if (result.nonEmpty) {
      push(out, result)
    }
    super.onUpstreamFinish()
  } else {
    pull(in)
  }
}

override def onUpstreamFinish(): Unit = {
  isFinishing = true
}
```
which works when the test case uses `Source(SampleElements.All)` (as in https://github.com/akka/akka-stream-contrib/pull/28#discussion_r69745163). However, the original test case (using `TestSource` and `sendNext`) fails now, because the last group is not pushed:
```
java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for OnNext(List(Element(6,3)))
```
Just out of curiosity, I'd like to know what's wrong with the latter (original) test case - can you please help me spot the problem?

patriknw: I would use `completeStage` instead of `super.onUpstreamFinish()`, but that is probably not the issue

patriknw: try
```scala
override def onUpstreamFinish(): Unit = {
  if (isAvailable(out)) {
    val result = buffer.result()
    push(out, result)
    super.onUpstreamFinish()
  } else
    isFinishing = true
}
```

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        currentState = Some(nextState)
+      }
+
+      override def onPull(): Unit = {
+        pull(in)
+      }
+
```
patriknw: add `buffer.clear()` in `postStop` to be nice to memory/gc

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
patriknw: an alternative, which is perhaps much more clean is to use `emit`. Search for "In some cases it is inconvenient and error prone to react on the regular state machine events" in http://doc.akka.io/docs/akka/2.4.7/scala/stream/stream-customize.html

Note that you should be able to do
```
override def onUpstreamFinish(): Unit = {
  emit(out, result)
  completeStage() // will be done after the emit has been pushed
}
```

and then no need for the flag

rucek: This is indeed much cleaner, thanks for pointing out this part of the docs - I somehow missed it.

rucek:
Are we ready to squash the commits? Or not yet?

rucek:
@patriknw anything else I should do here?

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -0,0 +1,91 @@
...
+    private var currentState: Option[Property] = None
+    private val buffer = Vector.newBuilder[Element]
+
+    setHandlers(in, out, new InHandler with OutHandler {
+
+      override def onPush(): Unit = {
+        val nextElement = grab(in)
+        val nextState = propertyExtractor(nextElement)
+
+        if (currentState.isEmpty || currentState.contains(nextState)) {
```
hochgi: condition can be replaced with: `currentState.forAll(_ == nextState)`

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -0,0 +1,91 @@
...
+          buffer += nextElement
+          pull(in)
+        } else {
+          val result = buffer.result()
+          buffer.clear()
+          buffer += nextElement
+          push(out, result)
+        }
+
+        currentState = Some(nextState)
```
hochgi: Theres no need to replace `currentState` for every pushed element. You can move this inside the else clause and save some unneeded option wrapping.

src/test/scala/akka/stream/contrib/AccumulateWhileUnchangedSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+} with AccumulateWhileUnchangedSpec
+
+trait AccumulateWhileUnchangedSpec extends BaseStreamSpec {
+
+  "AccumulateWhileUnchanged" should {
+
+    "emit accumulated elements when the given property changes" in {
+      val (_, sink) = Source(SampleElements.All)
+        .via(AccumulateWhileUnchanged(_.value))
+        .toMat(TestSink.probe)(Keep.both)
```
hochgi: You can use `Keep.right` and store as `val sink = ...` instead of `val (_, sink) = ...`

hochgi:
@rucek came here after reading your blog post (very nice post BTW), I added a few comments about some minor changes that cought my eyes while reading. Thanks for the post & PR :)

src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
rucek: Please note that `currentState` is initially `None`, so if I move the assignment into the `else` clause, I will still need to somehow initialize the `currentState` when the first element arrives. This in turn would require an additional `if` clause be evaluated for every element, so it's not cost-free either. So I'd stick to the current solution for the sake of readability.

hochgi: You are right... hmmm... So maybe a pattern match can help here?
```scala
currentState match {
  case None => ...
  case Some(s) if s == nextState => ...
  case _ => ...
}
```

rucek: With the pattern matching approach, some common logic would be required for the `None` and `Some(s)` parts (which could of course be extracted to a method), so we'd end up with something like:

```scala
def appendAndPull(): Unit = {
  buffer += nextElement
  pull(in)
}

currentState match {
  case None =>
    currentState = Some(nextState)
    appendAndPull()
  case Some(s) if s == nextState =>
    appendAndPull()
  case _ =>
    val result = buffer.result()
    buffer.clear()
    buffer += nextElement
    push(out, result)
    currentState = Some(nextState)
}
```
which is not necessarily better in my opinion (but indeed does less option wrapping).

As for the `currentState = currentState.orElse(Some(nextState))` - this is nothing more than an additional `if`, just hidden.

hochgi: but what you gain is not wrapping a value with option when it's not needed. the `.orElse` for `Some` simply returns `this`, so you always get a full option without generating garbage option instances, and and you can move the assignment of the `nextState` to the else clause :)

viktorklang: case Some(s) if s == nextState =>

==

case Some(`nextState`) =>

-- 
Cheers,
√

hochgi: @viktorklang rly?!?! that's awsome!
Thanks for the tip!

rucek: @hochgi so the bottomline of our discussion seems to be optimizing for memory (avoiding unnecessary `Option`s) vs. optimizing for CPU (avoiding an `if` which evaluates to `true` only for the first element). 

I'm really not sure which one is more important here - @patriknw @viktorklang any ideas?

drewhk: The additional "if" clause has basically zero cost, especially since it will be highly predictable by the CPU. I think Some.apply() is much more costly than that.

rucek: Makes sense, fixed.

johanandren:
LGTM

johanandren:
Thanks @rucek !

Issue #32, opened 2016-07-12
Add UnfoldFlow Source generator


hochgi:
a pattern I see comes up often in my code, is "unfolding" from flow.
I'll start with a pretty simple (yet useful) example:

consider Elasticsearch's [scroll API](https://www.elastic.co/guide/en/elasticsearch/guide/2.x/scroll.html). using it, you can traverse large amounts of data. So using the existing API, one might come up with something similar to (pseudo code - not real http client):

```scala
Source.unfoldAsync(initialScrollId) { scrollId =>
  val res: Future[HttpResponse] = httpClient.request(buildRequestFromId(scrollId))
  res.map { r =>
    val data = extractDataFrom(r)
    val next =  extractNextScrollIdFrom(r)
    if(data.isEmpty) None
    else Some(data -> next)
  }
}
```

well, since we are using streams, it might be much more efficient to use `akka-http`s connection pool flows. only problem, we don't want redundant sub streams materialization, we might be using with `Source.unfoldAsync`:

```scala
//BAD EXAMPLE
Source.single(buildRequestFromId(scrollId)).via(connectionPoolFlow).runWith(Sink.head)
```

So, I'm suggesting to create an unfolding source from a flow, by building a graph similar to:

![Source.unfoldFlow](http://i.imgur.com/MOmBWLJ.png)

API should be pretty straight forward:
```scala
def unfoldFlow[A,B,C,M](seed: A, flow: Flow[A,B,M])(unfoldWith: B => Option[(A,C)]): Source[C,M]
```

If you guys think it's useful, I'd be happy to create a PR for this.
and, IMO, this is something that could eventually migrate to the main streams module.

hochgi:
some concerns: 
- given flow might output more elements than what it takes for input.
- given flow might output less elements than what it takes for input.
- given flow might ask for more elements before emitting old ones.

possible handling of these cases:
- doesn't matter if more elements are coming out. we can decide the last element is the one used to create the next output elements. or the first received after a push to the flow, or we might generalize it and also accept a handling strategy.
- if less elements are outputted, than we can just re-send the last element again. if the flow will stay stuck, not emitting anything, we can consider this as the user fault. also, we might solve this with (an overkill solution IMHO) by getting a function `B => Option[(Source[A,_],C)]`, and emit different `A` elements to the flows when it asks for more.

hochgi:
A simpler signature would be:
```scala
[...]

PR #33, opened 2016-07-19
unfoldFlow & unfoldFlowWith Source generators added

hochgi:
PR for https://github.com/akka/akka-stream-contrib/issues/32
this is still a WIP. no tests & complete docs yet.
also, concerns raised in the issue need to be discussed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ target/
 ext-lib-src/
 .ensime
 .ensime_cache/
+moquette_store.mapdb
+moquette_store.mapdb.p
+moquette_store.mapdb.t
--- a/contrib/src/main/resources/reference.conf
+++ b/contrib/src/main/resources/reference.conf
@@ -1,0 +1,17 @@
+###################################
+# Akka Stream Contrib Config File #
+###################################
+
+# This is the reference config file that contains all the default settings.
+# Make your edits/overrides in your application.conf.
+
+akka {
+  stream {
+    contrib {
+      # Waiting time after inner flow provided by the user for unfold flow API cancels upstream,
+      # to get also the downstream cancelation (as graceful completion or failure which is propagated).
+      # If inner flow fails to complete/fail downstream, stage is failed with an IllegalStateException.
+      unfold-flow-timeout = 1m
+    }
+  }
+}
--- a/contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream._
+import akka.stream.scaladsl._
+import akka.stream.stage._
+import com.typesafe.config.ConfigFactory
+import scala.concurrent.duration.Duration
+
+/**
+ * Source factory methods are placed here
+ */
+object SourceGen {
+
+  /**
+   * Create a `Source` that will unfold a value of type `S` by
+   * passing it through a flow. The flow should emit a
+   * pair of the next state `S` and output elements of type `E`.
+   * Source completes when the flow completes.
+   *
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result,
+   * and may result in a deadlock.
+   */
+  def unfoldFlow[S, E, M](seed: S)(flow: Graph[FlowShape[S, (S, E)], M]): Source[E, M] = {
+
+    val generateUnfoldFlowGraphStageLogic = (shape: FanOutShape2[(S, E), S, E]) => new UnfoldFlowGraphStageLogic[(S, E), S, E](shape, seed) {
+      setHandler(nextElem, new InHandler {
+        override def onPush() = {
+          val (s, e) = grab(nextElem)
+          pending = s
+          push(output, e)
+          pushedToCycle = false
+        }
+      })
+    }
+
+    unfoldFlowGraph(new FanOut2unfoldingStage(generateUnfoldFlowGraphStageLogic), flow)
+  }
+
+  /**
+   * Create a `Source` that will unfold a value of type `S` by
+   * passing it through a flow. The flow should emit an output
+   * value of type `O`, that when fed to the unfolding function,
+   * generates a pair of the next state `S` and output elements of type `E`.
+   *
+   * IMPORTANT CAVEAT:
+   * The given flow must not change the number of elements passing through it (i.e. it should output
+   * exactly one element for every received element). Ignoring this, will have an unpredicted result,
+   * and may result in a deadlock.
+   */
+  def unfoldFlowWith[E, S, O, M](seed: S, flow: Graph[FlowShape[S, O], M])(unfoldWith: O => Option[(S, E)]): Source[E, M] = {
+
+    val generateUnfoldFlowGraphStageLogic = (shape: FanOutShape2[O, S, E]) => new UnfoldFlowGraphStageLogic[O, S, E](shape, seed) {
+      setHandler(nextElem, new InHandler {
+        override def onPush() = {
+          val o = grab(nextElem)
+          unfoldWith(o) match {
+            case None => completeStage()
+            case Some((s, e)) => {
+              pending = s
+              push(output, e)
+              pushedToCycle = false
+            }
+          }
+        }
+      })
+    }
+
+    unfoldFlowGraph(new FanOut2unfoldingStage(generateUnfoldFlowGraphStageLogic), flow)
+  }
+
+  /** INTERNAL API */
+  private[akka] def unfoldFlowGraph[E, S, O, M](
+    fanOut2Stage: GraphStage[FanOutShape2[O, S, E]],
+    flow:         Graph[FlowShape[S, O], M]
+  ): Source[E, M] = Source.fromGraph(GraphDSL.create(flow) {
+    implicit b =>
+      {
+        f =>
+          {
+            import GraphDSL.Implicits._
+
+            val fo2 = b.add(fanOut2Stage)
+            fo2.out0 ~> f ~> fo2.in
+            SourceShape(fo2.out1)
+          }
+      }
+  })
+}
--- a/contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
+++ b/contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
@@ -1,0 +1,218 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.scaladsl._
+import akka.stream.{ FlowShape, KillSwitches, OverflowStrategies }
+import akka.stream.testkit.scaladsl.TestSink
+
+class UnfoldFlowSpecAutoFusingOn extends { val autoFusing = true } with UnfoldFlowSpec
+class UnfoldFlowSpecAutoFusingOff extends { val autoFusing = false } with UnfoldFlowSpec
+
+trait UnfoldFlowSpec extends BaseStreamSpec {
+
+  "unfoldFlow" should {
+
+    "unfold Collatz conjecture with a sequence of 111 elements" should {
+
+      val done = new Exception("done")
+      val outputs = List(27, 82, 41, 124, 62, 31, 94, 47, 142, 71, 214, 107, 322, 161, 484, 242, 121, 364, 182, 91, 274, 137, 412, 206, 103, 310, 155, 466, 233, 700, 350, 175, 526, 263, 790, 395, 1186, 593, 1780, 890, 445, 1336, 668, 334, 167, 502, 251, 754, 377, 1132, 566, 283, 850, 425, 1276, 638, 319, 958, 479, 1438, 719, 2158, 1079, 3238, 1619, 4858, 2429, 7288, 3644, 1822, 911, 2734, 1367, 4102, 2051, 6154, 3077, 9232, 4616, 2308, 1154, 577, 1732, 866, 433, 1300, 650, 325, 976, 488, 244, 122, 61, 184, 92, 46, 23, 70, 35, 106, 53, 160, 80, 40, 20, 10, 5, 16, 8, 4, 2)
+
+      "with flow" in {
+
+        val source = SourceGen.unfoldFlow(27)(Flow.fromFunction[Int, (Int, Int)] {
+          case 1               => throw done
+          case n if n % 2 == 0 => (n / 2, n)
+          case n               => (n * 3 + 1, n)
+        }.recover {
+          case `done` => (1, 1)
+        })
+
+        val sink = source.runWith(TestSink.probe)
+
+        outputs.foreach { output =>
+          sink.request(1)
+          sink.expectNext(output)
+        }
+        sink.request(1)
+        sink.expectNext(1)
+        sink.expectComplete()
+      }
+
+      "with buffered flow" in {
+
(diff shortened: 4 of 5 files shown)


hochgi:
unfoldFlowWith seems overly complicated. The reason for this is that I'm already using this, and the flow I'm unfolding through, is an Http connection pool flow, and for better efficiency, it is computed eagerly.

even if downstream did not request the next element, the source could "pre-cook" a value that will be ready once it is pulled.

In second thought, it may not be wise, specifically for a generic library method. A better approach would be to pass the state through the flow lazily, when source is pulled. And in case one wants to compute values eagerly, a buffer can be used.

drewhk:
As usual, the stuff that can cause the most issues here is completion. I recommend adding tests that exercise various close event orderings (cancel/complete races, etc).

2m:
@hochgi can you move these two files under `contrib` directory? This is going to be their new home after we refactored this repo to multi-project (#39).

drewhk:
Ok, I am reviewing this now.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+ */
+object SourceGen {
+
+  /**
+   * Create a `Source` that will unfold a value of type `S` by
+   * passing it through a flow. The flow should emit a
+   * pair of the next state `S` and output elements of type `E`.
+   * Source completes when the flow completes.
+   */
+  def unfoldFlow[S, E, M](seed: S)(flow: Graph[FlowShape[S, (S, E)], M]): Source[E, M] = {
```
drewhk: Doesn't it make sense to implement this in terms of the other one with the Option? I know it adds some overhead, but there are allocations because of a Pair anyway and I don't think an extra Map is that costly. Or, at least benchmark the two options whether it is worth to keep them quasi-separate implementations.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+      }
+  })
+
+  private[akka] def fanOut2unfoldingStage[O, S, E](seed: S, withInHandler: ((S, E) => Unit, () => O, () => Unit) => InHandler) = new GraphStage[FanOutShape2[O, S, E]] {
+
+    override val shape = new FanOutShape2[O, S, E]("unfoldFlow")
+    override def createLogic(attributes: Attributes) = {
+
+      new GraphStageLogic(shape) {
+
```
drewhk: can you add two `val`s that give better names to the ports? Like out1-> output out2->feedback in->nextElem or something like that.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+        setHandler(out1, new OutHandler {
+          override def onPull() = {
+            pull(in)
+            if (!pushedToCycle && isAvailable(out0)) {
+              push(out0, pending)
+              pending = null.asInstanceOf[S]
+              pushedToCycle = true
+            }
+          }
+        })
```
drewhk: so downstream cancellation both completes and cancels the wrapped flow from both sides. That looks good.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+
+        setHandler(out0, new OutHandler {
+          override def onPull() = if (!pushedToCycle && isAvailable(out1)) {
+            push(out0, pending)
+            pending = null.asInstanceOf[S]
+            pushedToCycle = true
+          }
+
+          override def onDownstreamFinish() = {
+            //Do Nothing, intercept completion as downstream
```
drewhk: Hm, this means though that we can not serve any unfold outputs anymore, so if a next external pull comes then we won't be able to serve it and deadlock?

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+              case Some((s, e)) => handleStateElement(s, e)
+            }
+          }
+        }
+    })
+
+    unfoldFlowGraph(fanOut2Stage, flow)
+  }
+
+  private[akka] def unfoldFlowGraph[E, S, O, M](
```
drewhk: If we keep the dual-implementation way (i.e. not using the Option one to implement the other) then I would prefer subclassing to injecting this function. For me it reads better (even though I am not a big fan of subclassing).

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Source factory methods are placed here
+ */
+object SourceGen {
+
+  /**
+   * Create a `Source` that will unfold a value of type `S` by
+   * passing it through a flow. The flow should emit a
+   * pair of the next state `S` and output elements of type `E`.
+   * Source completes when the flow completes.
```
drewhk: Document in both cases that non 1-to-1 flows will simply deadlock this Source (i.e. no dropping or "multiplying" elements are supported).

contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+        }
+        sink.request(1)
+        sink.expectComplete()
+      }
+    }
+
+    "increment integers & handle KillSwitch" should {
+
+      "with simple flow" should {
+
```
drewhk: Add a test case where internal flow Cancels but does not complete, then downstream pulls.

drewhk:
Looks really great, only minor stuff, the main logic is sound and likely very useful. My only worry are non 1-to-1 flows as they can do very bad things but this is not the first place where this is an issue.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
hochgi: Not sure how a deadlock can manifest here.
we can simplify and deal with basically 3 states:

1. downstream pulled, but we haven't forward to cycle yet
2. there is an element in the cycle
3. idle (we already pushed downstream, and did not receive another demand yet)

in any of these 3 states, when inner flow decides to complete/abort, we should be notified twice. once as the upstream (which we ignore), and once as the downstream (which we propagate to downstream).

the "hard" scenario is:

- feedback becomes unavailable
- output is pulled -> inlet is pulled
- inlet becomes unavailable
- completion/error is propagated to downstream

but it's still OK.
or am I missing something here?

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
hochgi: benchmark code:
https://github.com/hochgi/akka-stream-contrib/commit/2bd97d6463b938b231a4147fe9d8e7636c7f6b70
with fusing, the overhead is rather small, but when fusing is disabled it is very much noticeable. some sample runs:
```none
seed = 1487492288

wrappedTime: 4.241016340000001 ms
bareTime: 2.6165949900000003 ms

wrappedTime: 3.141815640000001 ms
bareTime: 2.008202889999999 ms

seed = 27

wrappedTime: 2.0119849199999993 ms
bareTime: 1.3814176399999993 ms

wrappedTime: 4.65674252 ms
bareTime: 1.7444834599999999 ms
```

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
hochgi: Code readability is important, but for some reason, I like it better that way. The code is still pretty clear (IMO at least...), and I would prefer adding some comments explaining the injection than change the implementation.

Nevertheless, if you still think I should change implementation, I will.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
hochgi: done

contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
hochgi: done

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
drewhk: The scenario I was thinking about was:
 - no element in cycle
 - feedback becomes unavailable
 - output is pulled -> inlet is pulled
 - inlet does NOT become unavailable

I.e. the inner Flow only cancels. We can still cleanup in this case since we know that there is no element in the cycle so we will never be able to satisfy the downstream pull anyway, so we can just complete.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
drewhk: Well, then maybe I would prefer two subclasses instead of plugging in a factory function.

contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
```diff
@@ -0,0 +1,198 @@
...
+            sink.requestNext(1)
+            sink.requestNext(2)
+            sink.requestNext(3)
+            killSwitch.shutdown()
+            sink.expectComplete()
+
+          }
+        }
+      }
+
```
drewhk: Please also add a test case where the inner Flow has a buffer (i.e. just prepend or append a buffer stage). That exercises more paths.

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
hochgi: Oh... haven't thought of that. I still prefer to catch as downstream if I can, since I can propagate the completion properly (failure or finish). as upstream I cannot tell.

what do you suggest? maybe schedule a check if inlet became available after some time? what would be a proper time to wait (feels like it should be configurable, since it depends on the flow the user provides)

drewhk: I would not try to fix the generic case, only the case where we can be sure that we can't fulfill any further demand. I.e.:
 - feedback port is closed
 - no element in cycle
If the above two become true at the same time, then we can close as there will be no chance to emit any new element anymore. While this does not cover *all* cases, at least this is a reliable test.

hochgi: close, i.e complete gracefully.
so if inner flow was aborted (due to some exception), and I happen to get `feedback.onDownstreamFinish` before `nextElem.onUpstreamFailure(error)`, this will swallow the error.

drewhk: That's a good point. Hm, maybe the timeout behavior is the safest one.

contrib/src/test/scala/akka/stream/contrib/UnfoldFlowSpec.scala
hochgi: done

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,141 @@
...
+            import GraphDSL.Implicits._
+
+            val fo2 = b.add(fanOut2Stage)
+            fo2.out0 ~> f ~> fo2.in
+            SourceShape(fo2.out1)
+          }
+      }
+  })
+
+  private[akka] abstract class UnfoldFlowGraphStageLogic[O, S, E] private[stream] (shape: FanOutShape2[O, S, E], seed: S) extends GraphStageLogic(shape) {
```
hochgi: should I not restrict with `private[akka]`?
https://github.com/akka/akka/issues/20543

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,141 @@
...
+        if (!pushedToCycle && isAvailable(feedback)) {
+          push(feedback, pending)
+          pending = null.asInstanceOf[S]
+          pushedToCycle = true
+        }
+      }
+    })
+  }
+
+  private[akka] class FanOut2unfoldingStage[O, S, E] private[stream] (generateGraphStageLogic: FanOutShape2[O, S, E] => UnfoldFlowGraphStageLogic[O, S, E]) extends GraphStage[FanOutShape2[O, S, E]] {
```
hochgi: @drewhk should I not restrict with `private[akka]`/`private[stream]`?
https://github.com/akka/akka/issues/20543

drewhk: keep it restricted and slap a `/** INTERNAL API */` on them for javadoc. Thanks!

drewhk:
LGTM

hochgi:
squashed and pushed.
unsquashed commits can be found [here](https://github.com/hochgi/akka-stream-contrib/tree/32-hochgi-unfold-flow-unsquashed).

.gitignore
```diff
@@ -15,3 +15,6 @@ target/
 ext-lib-src/
 .ensime
 .ensime_cache/
+moquette_store.mapdb
+moquette_store.mapdb.p
+moquette_store.mapdb.t
```
johanandren: the what? :)

contrib/src/main/resources/reference.conf
```diff
@@ -0,0 +1,10 @@
+akka {
+  stream {
+    contrib {
+      # Waiting time after inner flow provided by the user for unfold flow API cancels upstream,
+      # to get also the downstream cancelation (as graceful completion or failure which is propagated).
+      # If inner flow fails to complete/fail downstream, stage is failed with an IllegalStateException.
+      unfold-flow-timeout = 1m
+    }
+  }
+}
```
johanandren: Maybe add a header block with "Akka Streams Contrib" and a short description like in the regular akka reference.conf files?

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -0,0 +1,144 @@
...
+
+            val fo2 = b.add(fanOut2Stage)
+            fo2.out0 ~> f ~> fo2.in
+            SourceShape(fo2.out1)
+          }
+      }
+  })
+
+  /** INTERNAL API */
+  private[akka] abstract class UnfoldFlowGraphStageLogic[O, S, E] private[stream] (shape: FanOutShape2[O, S, E], seed: S) extends GraphStageLogic(shape) {
```
johanandren: I think it would be nicer if put in a separate `UnfoldFlowGraphStageLogic` source file rather than in `SourceGen.scala`, it is meant to be a central factory place not the place for all custom graph stages.

johanandren: Uff. I meant a `UnfoldFlow.scala` ofc :) was distracted as I was writing that

johanandren:
LGTM except where the code is placed and some minor other things

.gitignore
hochgi: just trying to apply "boyscout" pattern... these are files generated during tests.
I cannot be the only one who's bothered by seeing untracked files, and running `git clean -f` after every test :)

johanandren: Shouldn't they go into `target/` if they are generated during tests? Who generates them? Weird. (Not saying it is up to you to fix)

johanandren: Ah it's from the mqtt connector. Will create a ticket.

hochgi: should I remove these from the .gitignore?

johanandren: Nah, you can let them stay there.

contrib/src/main/resources/reference.conf
johanandren: Great! Thanks

johanandren:
BTW no need to squash as we can do that through the github ui when we merge.

hochgi:
already squashed (& amended)
next time I'll keep the PR unsquashed ;)

Issue #35, opened 2016-08-04
Directory watch - added, changed, deleted files as events


johanandren:
Idea for a new source, I have one lying around on my computer since earlier and there was discussions on akka/akka about such a source.

I'm thinking `Source[(Path, ChangeType), SomeWayToCancel]` or possibly skip that `SomeWayToCancel` since it can be achieved by combining with a `KillSwitch`

joearasin:
It's a bit of a hack, but to play around with other pieces of the system and punt the innards of this to later, I used:

```
  val monitor = RxMonitor()
  val observable: Observable[EventAtPath] = monitor.observable
  val publisher = RxReactiveStreams.toPublisher(JavaConversions.toJavaObservable(observable))
  val eventSource: Source[EventAtPath, NotUsed] = Source.fromPublisher(publisher)
  // Run things 
  monitor.registerPath(ENTRY_MODIFY, "/path/to/whatever")
```
using https://github.com/lloydmeta/schwatcher with RxReactiveStreams

johanandren:
We also discussed pulling in https://github.com/pathikrit/better-files and use their actor based watch. But I think I'll stick to using the JDK provided `WatchService` (https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#newWatchService--). It's a pretty horrible interface, but it's 0-dependencies so that is a nice thing. It should be using native OS facilities on some platforms AFAIR.

PR #36, opened 2016-08-04
A JDK WatchService based directory changes source #35

johanandren:
Refs #35


src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+   * @see [[DirectoryChanges]] for details
+   */
+  def apply(directory: Path, pollInterval: FiniteDuration = 500.millis, maxBufferSize: Int = 200): Source[(Path, Change), NotUsed] =
+    Source.fromGraph(new DirectoryChanges(directory, pollInterval, maxBufferSize))
+
+  /**
+   * Java API
+   * @see [[DirectoryChanges]] for details
+   */
+  def forDirectory(directory: Path): Source[Pair[Path, DirectoryChanges.Change], NotUsed] = apply(directory).map(t => Pair.create(t._1, t._2))
```
johanandren: Not sure about these factories, but anyway, if they stay this needs to return javadsl.Source ofc.

johanandren: Should be `create` I guess.

joearasin:
Just gave this a spin -- it's great!

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler, TimerGraphStageLogic }
+import akka.japi.Pair
+import com.sun.nio.file.SensitivityWatchEventModifier
+
+import scala.collection.immutable.Queue
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
+object DirectoryChanges {
+  sealed trait Change
```
patriknw: `Changed`? since the others are in past tense

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+  import DirectoryChanges._
+  private val out: Outlet[(Path, Change)] = Outlet("DirectoryChanges.out")
+
+  override protected def initialAttributes: Attributes = defaultAttributes
+
+  override def shape: SourceShape[(Path, Change)] = SourceShape.of(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+    assert(Files.exists(directoryPath), s"The path: $directoryPath does not exist")
+    assert(Files.isDirectory(directoryPath), s"The path $directoryPath is not a directory")
```
patriknw: `require` instead of `assert`

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+        else {
+          poll()
+          if (buffer.nonEmpty) pushHead()
+          else schedulePoll()
+        }
+      }
+    })
+
+    def schedulePoll(): Unit = {
+      scheduleOnce(true, pollInterval)
```
patriknw: `true` as timer key?

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+        if (buffer.nonEmpty) pushHead()
+        else schedulePoll()
+      }
+    }
+
+    def poll(): Unit = {
+      try {
+        watchKey.pollEvents().asScala.foreach { eventt =>
+
+          val event = eventt.asInstanceOf[WatchEvent[Path]]
```
patriknw: `case event: WatchEvent[Path] @unchecked =>` ?

patriknw:
looking good

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
drewhk: hm, not sure, for me that reads weird.

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  override protected def initialAttributes: Attributes = defaultAttributes
+
+  override def shape: SourceShape[(Path, Change)] = SourceShape.of(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+    assert(Files.exists(directoryPath), s"The path: $directoryPath does not exist")
+    assert(Files.isDirectory(directoryPath), s"The path $directoryPath is not a directory")
+
+    private var buffer = Queue.empty[(Path, Change)]
```
drewhk: I don't like immutable.Queue. Can this be an array backed queue from JDK instead?

drewhk: Also, overflowstrategy maybe?

johanandren: Started out with the strategy, but the types are akka private. Sure can use JDK queue, you have some more motivation for it than not liking it though?

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
johanandren: Maybe the events should be `Modification`, `Creation` and `Deletion` instead?

src/main/scala/akka/stream/contrib/DirectoryChanges.scala
drewhk: Yes, performance. As for the strategy, I don't even think any other strategy than fail makes sense here, since dropped events result in out-of sync view. A downstream fused dropping buffer can be added by the user if he for some reason still prefers dropping.

patriknw:
LGTM

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,213 @@
...
+import akka.stream.stage.AbstractOutHandler;
+import akka.stream.stage.GraphStage;
+import akka.stream.stage.GraphStageLogic;
+import akka.stream.stage.TimerGraphStageLogic;
+import com.sun.nio.file.SensitivityWatchEventModifier;
+import scala.Tuple2;
+import scala.concurrent.duration.FiniteDuration;
+
+import java.nio.file.*;
+import static java.nio.file.StandardWatchEventKinds.*;
```
ktoso: static imports separately? :p

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+public class DirectoryChanges extends GraphStage<SourceShape<Pair<Path, DirectoryChanges.Change>>> {
+
+  public enum Change {
+    Modification,
+    Creation,
+    Deletion
+  }
+
+  private final static Attributes defaultAttributes = Attributes.name("DirectoryChanges");
```
ktoso: DEFAULT_ATTRIBUTES

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  @Override
+  public Attributes initialAttributes() {
+    return defaultAttributes;
+  }
+
+  @Override
+  public GraphStageLogic createLogic(Attributes inheritedAttributes) {
+    if(!Files.exists(directoryPath)) throw new IllegalArgumentException("The path: '" + directoryPath + "' does not exist");
```
ktoso: `if (`

ktoso: I'm giving the most useful feedback ever so far, am I not? 😉

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,213 @@
...
+
+    try {
+      return new TimerGraphStageLogic(shape) {
+        private final Queue<Pair<Path, Change>> buffer = new ArrayDeque<>();
+        private final WatchService service = directoryPath.getFileSystem().newWatchService();
+        private final WatchKey watchKey = directoryPath.register(
+          service,
+          new WatchEvent.Kind<?>[] { ENTRY_CREATE, ENTRY_MODIFY, ENTRY_DELETE, OVERFLOW },
+          // this is com.sun internal, but the service is useless on OSX without it
+          SensitivityWatchEventModifier.HIGH
```
ktoso: good

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,213 @@
...
+          scheduleOnce("poll", pollInterval);
+        }
+
+        private void doPoll() {
+          try {
+            for (WatchEvent<?> event: watchKey.pollEvents()) {
+              final WatchEvent.Kind<?> kind = event.kind();
+
+              if (OVERFLOW.equals(kind)) {
+                failStage(new RuntimeException("Overflow from watch service: '" + directoryPath + "'"));
```
ktoso: Add a comment explaining that technically one may choose to ignore the overflow instead - depends what you need.
No need to change impl, but add comment to explain for readers we're aware

ktoso:
LGTM

2m:
Needs a rebase.

src/main/java/akka/stream/contrib/DirectoryChanges.java
2m: Another approach could be to throw a more concrete exception here, which would enable user to decide what happens on overflow using stream supervision.

src/main/java/akka/stream/contrib/DirectoryChanges.java
```diff
@@ -0,0 +1,231 @@
...
+                @SuppressWarnings("unchecked")
+                final Path path = (Path) event.context();
+                final Path absolutePath = directoryPath.resolve(path);
+                final Change change = kindToChange(kind);
+
+                buffer.add(new Pair<>(absolutePath, change));
+                if (buffer.size() > maxBufferSize) {
+                  failStage(new RuntimeException("Max event buffer size " +
+                    maxBufferSize + " reached for $path"));
+                }
```
2m: I think this might limit the usability of this stage. We could make it more general, so that the stage would stop polling watch service when the buffer gets full. Then user can make a decision and by putting a buffer after this stage can decide what to do on overflow - backpressure, fail, drop, etc...

johanandren: "future improvements" ?

It only limits the usability for the use case "a directory which we change very very often". While the use case, drop a file or a few files in a directory to trigger processing will work fine.

johanandren: I mean, we only poll when there is demand.

johanandren: When there is demand and the buffer is empty, so the case when it fails here is when there are more changes on one poll result than fits in the buffer.

2m: Yea. Could be future improvement. Lets ticketify? We could have a `contributions-welcome` label here as well for easy to pick up tasks.

contrib/src/test/java/akka/stream/contrib/DirectoryChangesTest.java
```diff
@@ -0,0 +1,143 @@
...
+    final Path createdFile = Files.createFile(testDir.resolve("test1file1.sample"));
+
+    final Pair<Path, DirectoryChanges.Change> pair1 = probe.expectNext();
+    assertEquals(pair1.second(), DirectoryChanges.Change.Creation);
+    assertEquals(pair1.first(), createdFile);
+
+    Files.write(createdFile, "Some data".getBytes());
+
+    probe.request(1);
+    final Pair<Path, DirectoryChanges.Change> pair2 = probe.expectNext();
```
2m: Here and a bit below these two lines could be one: ``` ... pair2 = probe.requestNext();```

johanandren: Golfing eh?

2m:
LGTM. Squashy squashy and then mergy mergy.

Issue #34, opened 2016-07-22
Provide "tail -f" like FileIO Source


ktoso:
It would be fun to have a Source that behaves like "tail -f" so one can monitor a file for changes.

Discussion was here https://github.com/akka/akka/issues/21004#issuecomment-234520485

PR #38, opened 2016-08-06
"tail -f" source

johanandren:
Refs #35


johanandren:
Why aren't our regular file-source and sink done using NIO when it is this easy, or did I miss something?

src/main/scala/akka/stream/contrib/FileTailSource.scala
```diff
@@ -0,0 +1,84 @@
...
+            } else {
+              // hit end, try again in a while
+              scheduleOnce(NotUsed, pollingInterval)
+            }
+
+          case Failure(ex) =>
+            failStage(ex)
+        }
+
+        chunkHandler = new CompletionHandler[Integer, NotUsed] {
```
viktorklang: If you use the AsyncCallback as the attachment of the `read`, then you could use a single CompletionHandler instance as the single instance could be implemented as:

~~~scala
object TailCompletionHandler extends CompletionHandler[Integer, AsyncCallback[Try[Integer]]] {
  override def completed(result: Integer, attachment: AsyncCallback[Try[Integer]]): Unit = 
    attachment.invoke(Success(result))
   override def failed(ex: Throwable, attachment: NotUsed): Unit =
     attachment.invoke(Failure(ex))
}
~~~

hepin1989:
Why not make use of WatchService?

src/main/scala/akka/stream/contrib/FileTailSource.scala
```diff
@@ -0,0 +1,93 @@
...
+ */
+final class FileTailSource(path: Path, maxChunkSize: Int, startingPosition: Long, pollingInterval: FiniteDuration) extends GraphStage[SourceShape[ByteString]] {
+  import FileTailSource.TailCompletionHandler
+
+  val out = Outlet[ByteString]("FileTailSource.out")
+  override val shape: SourceShape[ByteString] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with OutHandler {
+      assert(Files.isReadable(path), s"path $path is not readable")
```
patriknw: avoid `assert` since it will stop the actor system if triggered, use `require` or `if-throw` instead

patriknw:
looking good

src/main/scala/akka/stream/contrib/FileTailSource.scala
```diff
@@ -0,0 +1,93 @@
...
+        chunkCallback = getAsyncCallback[Try[Integer]] {
+          case Success(readBytes) =>
+            if (readBytes > 0) {
+              buffer.flip()
+              push(out, ByteString.fromByteBuffer(buffer))
+              position += readBytes
+              buffer.clear()
+            } else {
+              // hit end, try again in a while
+              scheduleOnce(NotUsed, pollingInterval)
```
drewhk: Use a constant String instead as the key

drewhk:
Testing needs to be figured out somehow, but it looks great so far. Would be fun to combine with WebSockets or streaming Http responses.

johanandren:
Yea, I saw someone made a small play based command line tail-this-file-as-http a while ago, so the idea had struck me as well.

johanandren:
@hepin1989 The watch service is horribly slow on OSX which is what I develop on, that's why :)

drewhk:
You can also do some backoff based polling for good latency-polling overhead tradeoff.

src/main/java/akka/stream/contrib/FileTailSource.java
```diff
@@ -0,0 +1,174 @@
...
+   * Aborting the stage can be done by combining with a [[akka.stream.KillSwitch]]
+   *
+   * @param path a file path to tail
+   * @param maxChunkSize The max emitted size of the `ByteString`s
+   * @param startingPosition Offset into the file to start reading
+   * @param pollingInterval When the end has been reached, look for new content with this interval
+   */
+  public static akka.stream.scaladsl.Source<ByteString, NotUsed> apply(Path path, int maxChunkSize, long startingPosition, FiniteDuration pollingInterval) {
+    return create(path, maxChunkSize, startingPosition, pollingInterval).asScala();
+  }
```
johanandren: This doesn't really work. Maybe put a corresponding object in the Scala sources for it.

src/main/java/akka/stream/contrib/FileTailSource.java
```diff
@@ -0,0 +1,174 @@
...
+  @Override
+  public GraphStageLogic createLogic(Attributes inheritedAttributes) {
+    try {
+      if (!Files.exists(path)) throw new IllegalArgumentException("Path '" + path + "' does not exist");
+      if (Files.isDirectory(path)) throw new IllegalArgumentException("Path '" + path + "' cannot be tailed, it is a directory");
+      if (!Files.isReadable(path)) throw new IllegalArgumentException("No read permission for '" + path + "'");
+
+      return new TimerGraphStageLogic(shape) {
+        private final ByteBuffer buffer = ByteBuffer.allocate(maxChunkSize);
+        private final AsynchronousFileChannel channel = AsynchronousFileChannel.open(path, StandardOpenOption.READ);
```
johanandren: This runs the IO on "the default thread pool", not sure what kind of thread pool that is or if it is something to worry about.

src/main/java/akka/stream/contrib/FileTailSource.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private final Path path;
+  private final int maxChunkSize;
+  private final long startingPosition;
+  private final FiniteDuration pollingInterval;
+  private final Outlet<ByteString> out = Outlet.create("FileTailSource.out");
+  private final SourceShape<ByteString> shape = SourceShape.of(out);
+
+  // this is stateless, so can be shared among instances
+  private static final CompletionHandler<Integer, AsyncCallback<Try<Integer>>> completionHandler = new CompletionHandler<Integer, AsyncCallback<Try<Integer>>>() {
```
johanandren: Should idiomatic Java be `COMPLETION_HANDLER`?

johanandren:
Now rebased on top of multi-module project.

patriknw:
LGTM, squash before merge

contrib/src/main/java/akka/stream/contrib/FileTailSource.java
```diff
@@ -0,0 +1,174 @@
...
+import java.nio.channels.CompletionHandler;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
+/**
+ * Read the entire contents of a file, and then when the end is reached, keep reading
+ * newly appended data. Like the unix command `tail -f`.
+ *
+ * Aborting the stage can be done by combining with a [[akka.stream.KillSwitch]]
```
ktoso: why not expand the example to show how one would materialize a `Cancellable` such that one can cancel the source using that?

johanandren: Hmmm. I still lean towards simpler is better here. Especially as you can compose to achieve the same behavior.

2m:
LGTM

johanandren:
Refs #34 i mean

PR #40, opened 2016-08-11
Mqtt Source

2m:
Mqtt Akka Stream Source implementation which uses Async Paho Client under the hood.

### DIFF
--- a/mqtt/src/main/scala/Mqtt.scala
+++ b/mqtt/src/main/scala/Mqtt.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib.mqtt
+
+import akka.util.ByteString
+import akka.stream.stage._
+import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
+import scala.util._
+import scala.language.implicitConversions
+
+case class MqttSourceSettings(
+  connectionSettings: MqttConnectionSettings,
+  topics:             Map[String, Int]
+)
+
+case class MqttConnectionSettings(
+  broker:      String,
+  clientId:    String,
+  persistence: MqttClientPersistence
+)
+
+case class MqttMessage(topic: String, payload: ByteString)
+
+trait MqttConnectorLogic { this: GraphStageLogic =>
+
+  import MqttConnectorLogic._
+
+  def connectionSettings: MqttConnectionSettings
+  def onConnect: AsyncCallback[IMqttAsyncClient]
+  def onMessage: AsyncCallback[MqttMessage]
+  def onConnectionLost: AsyncCallback[Throwable]
+
+  final override def preStart(): Unit = {
+    val client = new MqttAsyncClient(
+      connectionSettings.broker,
+      connectionSettings.clientId,
+      connectionSettings.persistence
+    )
+
+    client.connect((), connectHandler(client))
+    client.setCallback(new MqttCallback {
+      def messageArrived(topic: String, message: PahoMqttMessage) =
+        onMessage.invoke(MqttMessage(topic, ByteString(message.getPayload)))
+
+      def deliveryComplete(token: IMqttDeliveryToken) =
+        println(s"Delivery complete $token")
+
+      def connectionLost(cause: Throwable) =
+        onConnectionLost.invoke(cause)
+    })
+  }
+
+  private val connectHandler: IMqttAsyncClient => Try[IMqttToken] => Unit = client => {
+    case Success(_)  => onConnect.invoke(client)
+    case Failure(ex) => onConnectionLost.invoke(ex)
+  }
+}
+
+object MqttConnectorLogic {
+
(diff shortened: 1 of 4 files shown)


shiv4nsh:
@2m : I have resolved the issue on #43 It was just the Patience parameter. Please review it thanks :)

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -0,0 +1,67 @@
...
+ */
+package akka.stream.contrib.mqtt
+
+import akka.util.ByteString
+import akka.stream.stage._
+import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
+import scala.util._
+import scala.language.implicitConversions
+
+case class MqttSourceSettings(
```
patriknw: `final` for all `case class`

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -0,0 +1,67 @@
...
+
+case class MqttConnectionSettings(
+  broker:      String,
+  clientId:    String,
+  persistence: MqttClientPersistence
+)
+
+case class MqttMessage(topic: String, payload: ByteString)
+
+trait MqttConnectorLogic { this: GraphStageLogic =>
```
patriknw: `private[mqtt]`?

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  private val connectHandler: IMqttAsyncClient => Try[IMqttToken] => Unit = client => {
+    case Success(_)  => onConnect.invoke(client)
+    case Failure(ex) => onConnectionLost.invoke(ex)
+  }
+}
+
+object MqttConnectorLogic {
+
+  implicit def funcToMqttActionListener(func: Try[IMqttToken] => Unit): IMqttActionListener = new IMqttActionListener {
```
patriknw: is this used only in the internal implementation?

patriknw:
looking good

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -0,0 +1,67 @@
...
+    client.setCallback(new MqttCallback {
+      def messageArrived(topic: String, message: PahoMqttMessage) =
+        onMessage.invoke(MqttMessage(topic, ByteString(message.getPayload)))
+
+      def deliveryComplete(token: IMqttDeliveryToken) =
+        println(s"Delivery complete $token")
+
+      def connectionLost(cause: Throwable) =
+        onConnectionLost.invoke(cause)
+    })
```
johanandren: Isn't there a race here, connect happens before `setCallback` or does the client not accept messages before callback is set?

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -0,0 +1,67 @@
...
+case class MqttMessage(topic: String, payload: ByteString)
+
+trait MqttConnectorLogic { this: GraphStageLogic =>
+
+  import MqttConnectorLogic._
+
+  def connectionSettings: MqttConnectionSettings
+  def onConnect: AsyncCallback[IMqttAsyncClient]
+  def onMessage: AsyncCallback[MqttMessage]
+  def onConnectionLost: AsyncCallback[Throwable]
```
johanandren: Maybe define the AsyncCallbacks in this trait instead, and have the concrete connector just implement the synchronous callbacks, so source would implement `onMessage(message: MqTTMessage)`?

mqtt/src/test/scala/MqttSourceSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  def publish(topic: String, payload: String)(implicit server: Server) = {
+    val msg = new PublishMessage()
+    msg.setPayload(ByteString(payload).toByteBuffer)
+    msg.setTopicName(topic)
+    msg.setQos(QOSType.MOST_ONE)
+    server.internalPublish(msg)
+  }
+
+  def withBroker(subscriptions: Map[String, Int])(test: MqttSourceSettings => Server => ActorSystem => Materializer => Any) = {
```
johanandren: A matter of taste but I'd have done this as a trait each test case implements instead to get rid of the boilerplate-curry-signature.

2m: Agreed. This is too verbose. Will move to to a trait. I think there was something on abstracting over implicit parameter lists in scala/dotty roadmap... :)

mqtt/src/main/scala/Mqtt.scala
2m: Yes. It will probably be used in Sink impl as well. WIll make it private[mqtt]

mqtt/src/main/scala/Mqtt.scala
2m: I think you are spot on. This might be the cause of racy failures on Travis.

mqtt/src/main/scala/Mqtt.scala
2m: Ok. A valid point. Source will have even less coupling to `paho` APIs.

PR #43, opened 2016-08-11
Refactored the MqttSpec and added the basic functionality

shiv4nsh:
Refactored the Mqtt spec so that the test may not be  broken on CI. And hence Resolved the issue in the PR #40


2m:
Great. Thanks for the fix!

PR #45, opened 2016-08-12
Address Mqtt source review feedback

2m:
* remove race condition between connect and message handler set
* final case classes
* make handlers abstract, instead of asyncCallback getters
* make internal implementation private to the package
* abstract over implicit argument lists

### DIFF
--- a/mqtt/src/main/scala/Mqtt.scala
+++ b/mqtt/src/main/scala/Mqtt.scala
@@ -9,27 +9,34 @@ import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
 import scala.util._
 import scala.language.implicitConversions
 
-case class MqttSourceSettings(
+final case class MqttSourceSettings(
   connectionSettings: MqttConnectionSettings,
   topics:             Map[String, Int]
 )
 
-case class MqttConnectionSettings(
+final case class MqttConnectionSettings(
   broker:      String,
   clientId:    String,
   persistence: MqttClientPersistence
 )
 
-case class MqttMessage(topic: String, payload: ByteString)
+final case class MqttMessage(topic: String, payload: ByteString)
 
-trait MqttConnectorLogic { this: GraphStageLogic =>
+/**
+ *  Internal API
+ */
+private[mqtt] trait MqttConnectorLogic { this: GraphStageLogic =>
 
   import MqttConnectorLogic._
 
   def connectionSettings: MqttConnectionSettings
-  def onConnect: AsyncCallback[IMqttAsyncClient]
-  def onMessage: AsyncCallback[MqttMessage]
-  def onConnectionLost: AsyncCallback[Throwable]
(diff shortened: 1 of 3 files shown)


mqtt/src/main/scala/Mqtt.scala
```diff
@@ -9,27 +9,31 @@ import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
...
   broker:      String,
   clientId:    String,
   persistence: MqttClientPersistence
 )
 
-case class MqttMessage(topic: String, payload: ByteString)
+final case class MqttMessage(topic: String, payload: ByteString)
 
-trait MqttConnectorLogic { this: GraphStageLogic =>
+private[mqtt] trait MqttConnectorLogic { this: GraphStageLogic =>
```
johanandren: `/** Internal API`

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -49,6 +52,7 @@ trait MqttConnectorLogic { this: GraphStageLogic =>
       def connectionLost(cause: Throwable) =
         onConnectionLost.invoke(cause)
     })
+    client.connect((), connectHandler(client))
```
johanandren: 👍

mqtt/src/test/scala/MqttSourceSpec.scala
```diff
@@ -76,10 +87,22 @@ class MqttSourceSpec extends WordSpec with Matchers with ScalaFutures {
     val server = new Server()
     server.startServer()
     try {
-      test(settings)(server)(sys)(mat)
+      test(FixtureParam(settings, server, sys, mat))
     } finally {
       server.stopServer()
     }
   }
```
johanandren: Who stops the `ActorSystem`?

johanandren:
Lookin' good!

mqtt/src/main/scala/Mqtt.scala
```diff
@@ -9,27 +9,31 @@ import org.eclipse.paho.client.mqttv3.{ MqttMessage => PahoMqttMessage, _ }
...
-  def onConnect: AsyncCallback[IMqttAsyncClient]
-  def onMessage: AsyncCallback[MqttMessage]
-  def onConnectionLost: AsyncCallback[Throwable]
+  def handleConnection(client: IMqttAsyncClient): Unit
+  def handleMessage(message: MqttMessage): Unit
+  def handleConnectionLost(ex: Throwable): Unit
+
+  def onConnect = getAsyncCallback[IMqttAsyncClient](handleConnection)
+  val onMessage = getAsyncCallback[MqttMessage](handleMessage)
+  val onConnectionLost = getAsyncCallback[Throwable](handleConnectionLost)
```
patriknw: It's fine to create it like this in the constructor, but you are not allowed to `invoke` it before `preStart`. That is a problem if you expose the callback to the external world (via materialized value).

PR #47, opened 2016-08-17
Backpressure and multiple materialzation

2m:
Add backpressure signalt to MQTT library by blocking library thread, when the internal buffer becomes full.

Also add support for multiple materialization by moving subscription future inside of create logic.

### DIFF
--- a/mqtt/src/main/scala/Mqtt.scala
+++ b/mqtt/src/main/scala/Mqtt.scala
@@ -31,6 +31,13 @@ private[mqtt] trait MqttConnectorLogic { this: GraphStageLogic =>
 
   def connectionSettings: MqttConnectionSettings
   def handleConnection(client: IMqttAsyncClient): Unit
+
+  /**
+   * Callback, that is called from the MQTT client thread before invoking
+   * message handler callback in the GraphStage context.
+   */
+  def beforeHandleMessage(): Unit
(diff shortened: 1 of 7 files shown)


mqtt/src/main/scala/Mqtt.scala
```diff
@@ -31,6 +31,13 @@ private[mqtt] trait MqttConnectorLogic { this: GraphStageLogic =>
 
   def connectionSettings: MqttConnectionSettings
   def handleConnection(client: IMqttAsyncClient): Unit
+
+  /**
+   * Callback, that is called from the MQTT client thread before invoking
+   * message handler callback in the GraphStage context.
+   */
+  def beforeHandleMessage(): Unit
```
johanandren: Hmm. This is only to be able to do the backpressure using the semaphore so it's only for the Source, the sink will not use it. If so it's a bit weird that it is in the common connector trait.

johanandren: When I think of it this entire trait is just for incoming rather than generic, right?

2m: Currently it is, yea. Will have to refactor when implementing Sink.

johanandren:
LGTM

PR #49, opened 2016-08-19
Use git to describe version

2m:


README.md
```diff
@@ -9,17 +9,23 @@ This project provides a home to Akka Streams add-ons which does not fit into the
 * it is an experiment or requires iterations with user feedback before including into the stable core module
 * it requires a faster release cycle
 
-Dependencies
-------------
+Artifacts
+---------
```
ktoso: true

build.sbt
```diff
@@ -1,6 +1,10 @@
 lazy val root = (project in file(".")).
-  aggregate(contrib, mqtt)
+  aggregate(contrib, mqtt, amqp).
+  enablePlugins(GitVersioning)
 
 lazy val contrib = project
 lazy val mqtt = project
 lazy val amqp = project
+
+git.useGitDescribe := true
```
ktoso: are we sure that's a good idea?
Perhaps needs a small guide how to release, because I don't remember how to properly release so it does not end up with weird hash id in version :)

2m: I can add a guide. But it is just creating a tag and then `sbt publishSigned`

ktoso: Please add such info to the README, I'll totally screw up the order or something ;)

2m: I have added instructions to `CONTRIBUTING.md`.

ktoso:
Thanks, LGTM :)

PR #51, opened 2016-08-25
Fix akka#21028 - Publish de Akka Stream TestKit utilities

cchantep:
See https://github.com/akka/akka/issues/21028

### DIFF
--- a/contrib/build.sbt
+++ b/contrib/build.sbt
@@ -4,7 +4,8 @@ lazy val contrib = (project in file(".")).
 name := "akka-stream-contrib"
 
 libraryDependencies ++= Seq(
-  "junit"            %  "junit"           % "4.12" % Test, // Common Public License 1.0
-  "com.novocode"     %  "junit-interface" % "0.11" % Test, // BSD-like
-  "com.google.jimfs" %  "jimfs"           % "1.1"  % Test  // ApacheV2
+  "com.typesafe.akka" %% "akka-stream-testkit" % Common.AkkaVersion,
+  "junit"             %  "junit"               % "4.12" % Test, // Common Public License 1.0
+  "com.novocode"      %  "junit-interface"     % "0.11" % Test, // BSD-like
+  "com.google.jimfs"  %  "jimfs"               % "1.1"  % Test  // ApacheV2
 )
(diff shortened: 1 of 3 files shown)


contrib/src/main/scala/akka/stream/contrib/TestKit.scala
```diff
@@ -0,0 +1,56 @@
...
+import akka.actor.ActorRef
+import akka.actor.ActorRefWithCell
+import akka.stream.Materializer
+import akka.stream.impl._
+import akka.testkit.TestProbe
+import com.typesafe.config.ConfigFactory
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+object TestKit {
```
cchantep: Not sure what's the best naming for this one

contrib/src/main/scala/akka/stream/contrib/TestKit.scala
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.constrib
```
johanandren: `constrib` -> `contrib`

johanandren:
I think we should take a step back here and consider if we should make the copied test tools a part of the stream testkit rather than duplicating them.

drewhk:
I agree. The Coroner is not streams stuff though but akka-actor-tests. I think for contrib it might be enough to have the assertAllStagesStopped to be exposed. But then again, it might be better to polish that functionality to a more general management facility for running streams. Maybe let's merge this here and remove it once we built better support for introspecting running streams?

contrib/build.sbt
```diff
@@ -4,7 +4,8 @@ lazy val contrib = (project in file(".")).
 name := "akka-stream-contrib"
 
 libraryDependencies ++= Seq(
-  "junit"            %  "junit"           % "4.12" % Test, // Common Public License 1.0
-  "com.novocode"     %  "junit-interface" % "0.11" % Test, // BSD-like
-  "com.google.jimfs" %  "jimfs"           % "1.1"  % Test  // ApacheV2
+  "com.typesafe.akka" %% "akka-stream-testkit" % Common.AkkaVersion,
```
cchantep: For now that's only used there

contrib/src/main/scala/akka/stream/contrib/TestKit.scala
cchantep: Fixed

drewhk:
So we opted for merging right?

johanandren:
Yes, LGTM!

cchantep:
Thanks

drewhk:
Thanks for your contribution!

Issue #50, opened 2016-08-25
FlowEnrichment stage


talp-traiana:
While working with akka-streams, I had a couple of flows where I had, for example, a (id, message) tuple but the ID was just carried along the stream. Everything in the stream itself (maps and filters, mostly) only cared about the message. The resulting code was pretty ugly (pattern matching the tuple at every point, with type annotations since the type-inference went haywire), and so I wrote a GraphStage that wraps a Flow and adds/removes the ID.

Do you think this is something that can be added here?

ktoso:
I'd be very happy to see that actually. Seen such patterns in big data things before but noone had a nice api for it. 

Interesting idea would be:

```scala
  Source.single("name" -> 42)
    .focusOn(_._2) 
      .map(_ + 58)      // could these be SubFlow/Source?
      .filter(_ > 10)   // could these be SubFlow/Source?
    .unfocus
    .map { case (name, num) => s"$name -> $num" }
```

talp-traiana:
I'll upload the code and open a PR later, so you can have a look.
The focus/unfocus looks nice, but I'm not sure how it would work. Where will it keep the "current" ID?

The current API looks like this:
```
    val innerFlow = Flow[String].map(_.toUpperCase).filterNot(s => s.startsWith("T"))
    val foo = Source(List((1l, "one"), (2l, "two"), (3l, "three"), (4l, "four")))
      .via(enrich(innerFlow))
      .runWith(Sink.foreach(println))
```

ktoso:
Perhaps exact syntax of focusOn is unachievable, but I think the feature should be possible.
Effectively `.map(el => f(ef))` is actually `.map(el => copy(f(extract(ef)))`.

We need the `copy` and `extract` implemented. Extract is trivial, it's just `_._2`, copy may be something we ask people to provide in `focusOn`, so `(t, s) => t.copy(bla = s)`.

It would be fun to see if we could use techniques from lenses here, see: https://github.com/julien-truffaut/Monocle

talp-traiana:
Ooh, good idea. I'll try to see if I can make it work.

(Just to be clear, you don't mean to actually use Monocle/catz/whatever, right?)

ktoso:
We could, if it turns out to be awesome might be very fun project.
Could live as separate project in akka-contrib-extra.

drewhk:
Issues are many-to-many Flows (this have been considered before). It might work in some cases but be aware of the issue.

[...]

PR #52, opened 2016-08-26
talpr-50-flow-tagger

talpr:
FlowTagger (issue #50)

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
@@ -1,5 +1,6 @@
 package akka.stream.contrib
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream._
 import akka.stream.contrib.FocusedSource.IncompleteFocusedSourceShape
@@ -10,8 +11,24 @@ import scala.annotation.unchecked.uncheckedVariance
 import scala.collection.immutable
 
 object FocusedFlow {
-  implicit class FocusedFlowOps[Tag, Src, Mat](from: Flow[_, (Tag, Src), Mat]) {
-    def focus = new FocusedFlow(BidiFlow.fromFlowsMat(from, Flow[Src])(Keep.left))
+  implicit class FocusedFlowOps[Tag, Src, In, Mat](from: Flow[Src, (Tag, In), Mat]) {
+    def focus = new FocusedFlow(BidiFlow.fromFlowsMat(from, Flow[In])(Keep.left))
+  }
+
+  def main(args: Array[String]): Unit = {
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+
+    val flow: Flow[(Int, String), (Int, String), NotUsed] = Flow[(Int, String)]
+      .focus
+        .map(_.toUpperCase)
+        .filterNot(_.startsWith("T"))
+      .unfocus
+    val foo = Source(List((1, "one"), (2, "two"), (3, "three"), (4, "four")))
+      .via(flow)
+      .runWith(Sink.foreach(println))
+
+    system.terminate()
   }
 }
 
@@ -103,7 +120,7 @@ private object FocusedSource {
 
   private final case class IncompleteFocusedSourceShape[+SrcIn, In, +Out](
    srcIn: Outlet[SrcIn@uncheckedVariance],
-   in: Inlet[In@uncheckedVariance],
+   in: Inlet[In],
    out: Outlet[Out@uncheckedVariance]
   ) extends Shape {
     override val inlets: immutable.Seq[Inlet[_]] = List(in)
--- a/contrib/src/main/scala/akka/stream/contrib/FlowTagger.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/FlowTagger.scala
@@ -1,8 +1,9 @@
 package akka.stream.contrib
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream._
-import akka.stream.scaladsl.{Flow, FlowOpsMat, GraphDSL, Keep, RunnableGraph, Sink, Source}
+import akka.stream.scaladsl.{Flow, FlowOps, GraphDSL, RunnableGraph, Sink, Source}
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 
 import scala.annotation.unchecked.uncheckedVariance
@@ -96,7 +97,7 @@ object FocusedFlow {
     val foo = Source(List((1l, "one"), (2l, "two"), (3l, "three"), (4l, "four")))
       .focus
         .map(_.toUpperCase)
-        .filterNot(_.startsWith("O"))
+        .filterNot(_.startsWith("T"))
       .unfocus
       .runWith(Sink.foreach(println))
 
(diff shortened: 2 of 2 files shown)


contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
```diff
@@ -0,0 +1,241 @@
...
+  val outTup = Outlet[(E, Out)]("out-tuple")
+  val toFlow = Outlet[In]("to-flow")
+  val fromFlow = Inlet[Out]("from-flow")
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var curr: E = _
+    setHandler(inTup, new InHandler {
+      override def onPush(): Unit = {
+        val (cur, in) = grab(inTup)
+        curr = cur
```
ktoso: Let's make it passed in which value we pick :-)

ktoso: Actually going with raw, sbt-boilerplate generated ones would be enough it seems,
and it also seems we do need a custom type, SubFlow won't do hm.

General idea would be:

```
  // need to generate for 22-arity
  implicit final class Focus2OnSource[+T1, +T2, +M](s: Source[(T1, T2), M]) {
  import TupleGet._
    def focus[T, TT1, TT2]( // fix variance
      get: Tuple2[T1, T2] => T, 
      set: Tuple2[T1, T2] => T => Tuple2[TT1, TT2]): impl.SubFlowImpl[T, T, M, scaladsl.Source[Out, M]#Repr, scaladsl.RunnableGraph[M]] = ???
  }

  Source.single("" -> 12)
    .focus(get = t => t._1, 
           set = t => v => t.copy(_1 = v))
```

And instead of the SubFlow we'd need a custom type there it seems though.
Would you be able to give it a shot towards n-arity? Then it would be really awesome and useful I think

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
```diff
@@ -0,0 +1,241 @@
...
+
+  def main(args: Array[String]): Unit = {
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+
+    val flow: Flow[(Int, String), (Int, String), NotUsed] = Flow[(Int, String)]
+      .focus
+        .map(_.toUpperCase)
+        .filterNot(_.startsWith("T"))
+      .unfocus
```
ktoso: I think this is really cool, though we need to solve [2-22]-arity

ktoso:
Very fun, hope to see the full thing soon :)

drewhk:
On the side: one fun idea in a similar vein would be to accept Lenses to "project-down" a stream. Then you would be able to handle any n-to-n streams, filtering, mapConcatting or whatever on types, given that there is a Lens for it.

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
```diff
@@ -0,0 +1,241 @@
...
+
+import scala.annotation.unchecked.uncheckedVariance
+import scala.collection.immutable
+
+object FocusedFlow {
+  implicit class FocusedFlowOps[Tag, Src, In, Mat](from: Flow[Src, (Tag, In), Mat]) {
+    def focus = new FocusedFlow(BidiFlow.fromFlowsMat(from, Flow[In])(Keep.left))
+  }
+
+  def main(args: Array[String]): Unit = {
```
drewhk: what about this `main` method? We would need unit tests instead :)

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
```diff
@@ -0,0 +1,241 @@
...
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+
+    val flow: Flow[(Int, String), (Int, String), NotUsed] = Flow[(Int, String)]
+      .focus
+        .map(_.toUpperCase)
+        .filterNot(_.startsWith("T"))
```
drewhk: will `mapConcat` or `mapAsync` work?

drewhk:
Hm, this approach seems to be dangerous to me. It works for filter, but I don't see how it can work for 1-to-n stages. Also, isn't this racy? I mean the unfocus part and the tag it applies.

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
talpr: @ktoso
So this brings up a couple of related questions:
- If the user has to provide the 'get' and 'set' mappings, then why limit it to tuples?
- If it is limited to tuples, then wouldn't it be better to generate methods like focus_1, focus_2,... ?
- What happens if the output of the flow is a different type from the input?

Originally I thought about having the user to provide the functions when calling focus/unfocus, but that made the API really cluttered and I don't think it saved too much (considering that the user can just add a 'map' themselves).

So what I'm trying to say is, I'll gladly try to make it work for larger tuples, but I'm not sure whether it really adds much.

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
talpr: Of course, this is still a draft... ^_^

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
drewhk: This is why I mentioned Lenses. They are basically reified getters/setters for immutable data. While it probably needs some good lense library (I think shapeless have them) they could work nicely. It is Scala only of course.

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
talpr: Hmm, interesting question. I guess that would depend on your definition of 'work', but most probably not in the way you (or the user) intends.

For mapAsync, assuming parallelism > 1 it won't work as expected. For mapConcat, it could work, assuming it doesn't buffer internally.

talpr:
@drewhk 
Indeed, it won't always work. 1-to-n is OK (assuming you expect them all to get the same tag), but anything that has internal buffers or anything happening out-of-order won't work.

Like I said in the original issue, this won't always work, but in many simple use-cases it can be very handy.

drewhk:
I just investigate if we can make it work better. For example if not only via is overridden, but mapAsync to automatically remove then add the tag, it can work, too.

talpr:
Yeah, that could work. It would probably require unfocusing the stream first, doing the mapAsync, and focusing again, but I guess that wouldn't be that bad...

contrib/src/main/scala/akka/stream/contrib/FocusedFlow.scala
jroper: Another question, will `grouped` work, and what behaviour would you expect from it?

Use case: Lagom offers Kafka streams as ordinary streams of messages with no envelope, the user supplies a `Flow[T, Done, _]` to handle it, and the emission of each `Done` causes the messages to be committed (this itself may be batched, but that's irrelevant here).  So Lagom would do this:

```scala
kafkaConsumer
  .focus
  .via(userSuppliedFlow)
  .unfocus
  .map(commit)
```

Now, if the user wants to batch processing themselves, for efficiency reasons, then they might use `Flow.groupedWithin` followed by a `mapAsync`. In the best case scenario, we could allow them to return a single `Done` for the batch, and the unfocus would just select either the first or the last message to associate that `Done` with. Alternatively, the user might have to do a `mapConcat` to ensure that they still emit one message for each message that was grouped.

jroper: See https://github.com/akka/akka/issues/22975 for more info.

hochgi: This seems like a common problem whenever an API takes some flow as input without knowing anything on it.
I had similar problems in:
https://github.com/akka/akka-stream-contrib/pull/33
https://github.com/akka/akka-stream-contrib/pull/26
Input flow may buffer, expand or reduce elements, it may also complete early (and asymmetrically, e.g: only completing upstream, but not downstream...).
I don't think it's possible to ensure anything on the flow.
In mentioned PRs, I left most of these worries as the user responsibility,
and added this: https://github.com/akka/akka-stream-contrib/blob/master/contrib/src/main/resources/reference.conf
to ensure a better completion for the stream.
but everything else is left at the user responsibility, and documented with warning against misuse.
I wonder if it would be possible to add a more constrained type of flow,
something like `Unbuffered1To1RationFlow[I,O,M] <: Flow[I,O,M]` (with a better name ;)).
but that's probably out of scope here, and much harder to implement...

ennru:
This hasn't seen any activity for a very long time.
I'll close it, please re-open or start over.

PR #53, opened 2016-08-29
Add Cassandra Source

atamborrino:
This PR provides a way to transform a Cassandra's ResultSet into a Akka Stream's Source, i.e. it lets you stream the result of a Cassandra query.

Example:

```scala
import akka.actor.ActorSystem
import akka.actor.stream.ActorMaterializer
import akka.stream.contrib.cassandra.CassandraSource
import com.datastax.driver.core.{SimpleStatement, Cluster}

implicit val system = ActorSystem()
implicit val mat = ActorMaterializer()

val cluster = Cluster.builder.addContactPoint("127.0.0.1").withPort(9042).build
implicit val session = cluster.connect()

val stmt = new SimpleStatement("SELECT * FROM akka_stream_test.test")

val cassSource: Source[Row, NotUsed] = CassandraSource(stmt)
```

### DIFF
--- a/cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
+++ b/cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib.cassandra
+
+import akka.NotUsed
+import akka.actor.ActorSystem
+import akka.stream._
+import akka.stream.scaladsl.Source
+import akka.stream.stage.{ AsyncCallback, OutHandler, GraphStageLogic, GraphStage }
+import com.datastax.driver.core.{ ResultSet, Session, Statement, Row }
+import com.google.common.util.concurrent.{ ListenableFuture, FutureCallback, Futures }
+import scala.concurrent.{ Future, Promise }
+import scala.collection.JavaConverters._
+import scala.util.{ Try, Failure, Success }
+
+class CassandraSource(futStmt: Future[Statement], session: Session) extends GraphStage[SourceShape[Row]] {
+  val out: Outlet[Row] = Outlet("CassandraSource.out")
+  override val shape: SourceShape[Row] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+      var maybeRs = Option.empty[ResultSet]
+      var futFetchedCallback: AsyncCallback[Try[ResultSet]] = _
+
+      override def preStart(): Unit = {
+        implicit val ec = materializer.executionContext
+
+        futFetchedCallback = getAsyncCallback[Try[ResultSet]](tryPushAfterFetch)
+
+        val futRs = futStmt.flatMap(stmt => guavaFutToScalaFut(session.executeAsync(stmt)))
+        futRs.onComplete(futFetchedCallback.invoke)
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit = {
+          implicit val ec = materializer.executionContext
+
+          maybeRs match {
+            case Some(rs) if rs.getAvailableWithoutFetching > 0 => push(out, rs.one())
+            case Some(rs) if rs.isExhausted                     => completeStage()
+            case Some(rs) =>
+              // fetch next page
+              val futRs = guavaFutToScalaFut(rs.fetchMoreResults())
+              futRs.onComplete(futFetchedCallback.invoke)
+            case None => () // doing nothing, waiting for futRs in preStart() to be completed
+          }
+        }
+      })
+
+      private def tryPushAfterFetch(rsOrFailure: Try[ResultSet]): Unit = rsOrFailure match {
+        case Success(rs) =>
+          maybeRs = Some(rs)
+          if (rs.getAvailableWithoutFetching > 0) {
+            if (isAvailable(out)) {
+              push(out, rs.one())
+            }
+          } else {
+            completeStage()
+          }
+
+        case Failure(failure) => failStage(failure)
+      }
+    }
+
+  private def guavaFutToScalaFut[A](guavaFut: ListenableFuture[A]): Future[A] = {
+    val p = Promise[A]()
+    val callback = new FutureCallback[A] {
+      override def onSuccess(a: A): Unit = p.success(a)
+      override def onFailure(err: Throwable): Unit = p.failure(err)
+    }
+    Futures.addCallback(guavaFut, callback)
+    p.future
+  }
+}
+
+object CassandraSource {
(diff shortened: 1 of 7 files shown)


atamborrino:
Tests currently need a local Cassandra node to run. I also have a version using cassandra-unit (https://github.com/jsevellec/cassandra-unit) to spawn an in-memory Cassandra, but it does not handle shutdown properly, meaning you can not run the tests twice in a same sbt session...

What do you think is best?

patriknw:
We should not include cassandra-unit, sine it's LGPL. I don't want to get close to that even if it would only be for tests.

atamborrino:
Ok. I just added Cassandra to travis ci.

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
```diff
@@ -0,0 +1,106 @@
...
+import akka.stream.scaladsl.Source
+import akka.stream.stage.{ OutHandler, GraphStageLogic, GraphStage }
+import com.datastax.driver.core.{ ResultSet, Session, Statement, Row }
+import com.google.common.util.concurrent.{ ListenableFuture, FutureCallback, Futures }
+import scala.collection.mutable
+import scala.concurrent.{ Future, Promise }
+import scala.collection.JavaConverters._
+import scala.util.{ Try, Failure, Success }
+
+class CassandraSource(stmt: Statement, session: Session) extends GraphStage[SourceShape[Row]] {
```
patriknw: we can compare with this other implementation: https://github.com/lagom/lagom/blob/master/persistence/src/main/scala/com/lightbend/lagom/internal/persistence/cassandra/CassandraSessionImpl.scala#L272

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
```diff
@@ -0,0 +1,106 @@
...
+      var maybeRs = Option.empty[ResultSet]
+
+      override def preStart(): Unit = {
+        implicit val ec = materializer.executionContext
+
+        val futRs = guavaFutToScalaFut(session.executeAsync(stmt))
+        val futRsCallback = getAsyncCallback[Try[ResultSet]] {
+          case Success(rs) =>
+            maybeRs = Some(rs)
+            queue.enqueue(rs.iterator().asScala.take(rs.getAvailableWithoutFetching).toSeq: _*)
```
drewhk: you can use the `emit` API for this.

patriknw: good suggestion, but I don't think there is a need for a buffer (queue) at all, see https://github.com/lagom/lagom/blob/master/persistence/src/main/scala/com/lightbend/lagom/internal/persistence/cassandra/CassandraSessionImpl.scala#L272

atamborrino:
Indeed using resultSet.one() there is no need for an internal buffer (the resultSet is itself the internal buffer). I've updated the code.

hochgi:
@atamborrino Hi, I haven't gotten too much into details, but since I'm using cassandra with streams, I wanted to ask what benefits does this implementation gives me over:

```scala
import scala.collection.JavaConverters._
Source.fromIterator(() => rs.iterator().asScala)
```
or
```scala
Source.fromIterator(() => new Iterator[Row] {
  override def hasNext: Boolean = rs.isExhausted
  override def next(): Row = rs.one
})
```

also, you might find [this sbt plugin](https://github.com/hochgi/sbt-cassandra) useful.
(although it is quite outdated. It should have long due to make it an `AutoPlugin`, but still, it's useful and serves us well)

patriknw:
@hochgi the first one is blocking, when it reaches end of pre-fetched rows. Second doesn't iterate over all rows, only the pre-fetched.

patriknw:
Perhaps isExhausted is end of result set, but then second has same blocking problem as the first

hochgi:
@patriknw thanks. So I may switch to this when PR is accepted :D
BTW, a blocking source is not too bad, if followed by a sufficient buffer.

patriknw:
Blocking is always bad, then you have to run it on a dedicated dispatcher. Anyway, this is solvable as you can see here.

atamborrino:
Ready for final review @patriknw

patriknw:
thanks, I'll try to review more next week

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
patriknw: When I implemented the other `SelectSource` I found it useful to have the `Statement` param as `Future[Statement]`. That makes it easier to compose with other async things like `prepareAsync`.

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
```diff
@@ -0,0 +1,88 @@
...
+      })
+
+      private def tryPushAfterFetch(rsOrFailure: Try[ResultSet]): Unit = rsOrFailure match {
+        case Success(rs) =>
+          maybeRs = Some(rs)
+          if (rs.getAvailableWithoutFetching > 0) {
+            if (isAvailable(out)) {
+              push(out, rs.one())
+            }
+          } else {
```
patriknw: ok, so we don't need to check `isExhausted` because should always have at least one row

patriknw:
LGTM, great contribution
Possibly change the param to `Future[Statement]` to encourage `prepareAsync` and friends.

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
```diff
@@ -0,0 +1,88 @@
...
+import akka.stream.scaladsl.Source
+import akka.stream.stage.{ AsyncCallback, OutHandler, GraphStageLogic, GraphStage }
+import com.datastax.driver.core.{ ResultSet, Session, Statement, Row }
+import com.google.common.util.concurrent.{ ListenableFuture, FutureCallback, Futures }
+import scala.concurrent.{ Future, Promise }
+import scala.collection.JavaConverters._
+import scala.util.{ Try, Failure, Success }
+
+class CassandraSource(stmt: Statement, session: Session) extends GraphStage[SourceShape[Row]] {
+  val out: Outlet[Row] = Outlet("CassandraSource")
```
ktoso: the name should be `CassandraSource.out`

cassandra/src/main/scala/akka/stream/contrib/cassandra/CassandraSource.scala
atamborrino: yes that's right

atamborrino:
I've changed the CassandraSource param to Future[Statement] and added the corresponding constructors. I think this is ready to be merged =)

patriknw:
LGTM, thanks for a great contribution @atamborrino 
I'll make a release right away.

PR #60, opened 2016-09-07
Add Debouncing

joearasin:
I'm not even sure if debounce is the proper name for this. The basic functionality is to aggregate elements according to a particular key, and require them to be older than a specified duration in order to be passed downstream. There's probably room for a better data structure as well.

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/Debounce.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/Debounce.scala
@@ -1,0 +1,132 @@
+package akka.stream.contrib
+
+import akka.stream._
+import akka.stream.stage._
+
+import scala.collection.mutable
+import scala.concurrent.duration._
+
+private sealed trait QueueResult[+T]
+private case class QueueElement[T](element: T) extends QueueResult[T]
+private case class NotReadyFor(duration: FiniteDuration) extends QueueResult[Nothing]
+private case object QueueEmpty extends QueueResult[Nothing]
+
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @joearasin,

raboof:
@joearasin hmm, do I understand correctly that as long as 'upstream' is sending elements for a key, this stage will never emit? In what kinds of cases would that be useful?

Could something like this be achieved by combining [partition](https://doc.akka.io/docs/akka/2.5/scala/stream/stages-overview.html#partition), [conflate](https://doc.akka.io/docs/akka/2.5/scala/stream/stages-overview.html#conflate), [throttle](https://doc.akka.io/docs/akka/2.5/scala/stream/stages-overview.html#throttle) and [merge](https://doc.akka.io/docs/akka/2.5/scala/stream/stages-overview.html#merge)?

joearasin:
Yes -- this will not emit as long as upstream is sending elements -- I'm trying to remember the context for this -- iirc, the usage was to put it after something like https://github.com/akka/alpakka/blob/master/file/src/main/scala/akka/stream/alpakka/file/scaladsl/DirectoryChangesSource.scala -- where the goal is to provide a bit of time for a file to stabilize before allowing downstream events to trigger.

raboof:
Closing this for now as there hasn't been a lot of activity on this ticket. Happy to reopen if there is interest, but then we'd have to make sure we add a nice description to the docs of how this stage behaves and in what kind of situations it is useful - and perhaps reconsider naming.

PR #63, opened 2016-09-13
Always releases backpressure semaphore

### LINKED ISSUE #62: MqttSource doesn't release backpressure semaphore for fast downstream
It acquires backpressure semaphore at every message arrival, but it only releases when it picks a message from queue. It doesn't release it when working with fast downstream.

tkawachi:
Fixes #62


drewhk:
Yep, great catch!!

2m:
Thanks. LGTM!

Issue #30, opened 2016-07-07
Do we need Accumulate


hseeberger:
According to the docs (I wrote): "This stage emits folded values like `scan`, but the first element emitted is not the zero value but the result of applying the given function to the given zero value and the first pushed element."

The same can be achieved with `scan(...)(...).drop(1)`.

What do others thing?

hseeberger:
@patriknw, @ktoso, (other Akka team members),
I suggest we get rid of this beautiful piece of code again, because it's simpler to use the two well known standard stages `scan` and `drop` in combination instead of introducing another one which is very similar to `scan`.
If you agree, I'd create a PR which deprecates `Accumulate`.

hseeberger:
@patriknw, @ktoso, @drewhk, any opinions?

drewhk:
The custom stage is more efficient, however it does not show good example to users (build from existing ops), so maybe better removing it.

johanandren:
I agree with deprecation and mention scan+drop as a replacement.

PR #66, opened 2016-10-01
Deprecate Accumulate (#30)

hseeberger:


2m:
We should deprecate `apply`/`create` methods instead.

johanandren:
Good point, that's the actual public API I guess.

hseeberger:
All right, added deprecation annotations to the factory methods. Please review again.

PR #73, opened 2016-10-14
Valve to control a flow (open/close)

regis-leray:
- implements a valve to control a flow through a switch

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/Valve.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/Valve.scala
@@ -1,0 +1,131 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler }
+import akka.stream._
+import akka.stream.contrib.SwitchMode.{ Close, Open }
+
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * Pause/ Resume a Flow
+ */
+sealed trait ValveSwitch {
+  /**
+   * Change the state of the valve
+   *
+   * @param mode expected mode to switch on
+   * @return A future that completes with true if the mode did change and false if it already was in the requested mode
+   */
+  def flip(mode: SwitchMode): Future[Boolean]
+}
+
+object Valve {
+  /**
+   * Factory for [[Valve]] instances.
+   *
+   */
+  def apply[A](): Valve[A] = Valve[A](SwitchMode.Open)
+
+  /**
+   * Java API: Factory for [[Valve]] instances.
+   *
+   */
+  def create[A](): Valve[A] = Valve[A](SwitchMode.Open)
+
+  /**
+   * Factory for [[Valve]] instances.
+   *
+   */
+  def apply[A](mode: SwitchMode): Valve[A] = new Valve[A](mode)
+
+  /**
+   * Java API: Factory for [[Valve]] instances.
+   *
+   */
+  def create[A](mode: SwitchMode): Valve[A] = Valve[A](mode)
+
+}
+
+/**
+ * Materializes into a [[ValveSwitch]] which provides a the method flip that stops or restarts the flow of elements passing through the stage. As long as the valve is closed it will backpressure.
+ *
+ * Note that closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost.
+ *
+ * @param mode state of the valve at the startup of the flow (by default Open)
+ */
+final class Valve[A](mode: SwitchMode) extends GraphStageWithMaterializedValue[FlowShape[A, A], ValveSwitch] {
+
+  val in: Inlet[A] = Inlet[A]("valve.in")
+
+  val out: Outlet[A] = Outlet[A]("valve.out")
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, ValveSwitch) = {
+    val logic = new ValveGraphStageLogic(shape, mode)
+    (logic, logic.switch)
+  }
+
+  private class ValveGraphStageLogic(shape: Shape, var mode: SwitchMode) extends GraphStageLogic(shape) with InHandler with OutHandler {
+    val switch = new ValveSwitch {
+
+      val callback = getAsyncCallback[(SwitchMode, Promise[Boolean])] {
+        case (flipToMode, promise) =>
+          val succeed = mode match {
+            case _ if flipToMode == mode => false
+
+            case Open =>
+              mode = SwitchMode.Close
+              true
+
+            case Close =>
+              if(isAvailable(in)) {
+                push(out, grab(in))
+              }else if (isAvailable(out)) {
+                pull(in)
+              }
+
+              mode = SwitchMode.Open
+              true
+          }
+
+          promise.success(succeed)
+      }
+
+      override def flip(flipToMode: SwitchMode): Future[Boolean] = {
+        val promise = Promise[Boolean]()
+        callback.invoke((flipToMode, promise))
+        promise.future
+      }
+    }
+
+    setHandlers(in, out, this)
+
+    override def onPush(): Unit = {
+      if (isOpen) {
+        push(out, grab(in))
+      }
+    }
+
+    override def onPull(): Unit = {
+      if (isOpen) {
+        pull(in)
+      }
+    }
+
+    private def isOpen = mode == SwitchMode.Open
+  }
+
+}
+
+trait SwitchMode
+
+object SwitchMode {
+
+  case object Open extends SwitchMode
+
+  case object Close extends SwitchMode
+}
--- a/contrib/src/test/scala/akka/stream/contrib/ValveSpec.scala
+++ b/contrib/src/test/scala/akka/stream/contrib/ValveSpec.scala
@@ -1,0 +1,232 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.actor.ActorSystem
+import akka.pattern.after
+import akka.stream.ActorMaterializer
+import akka.stream.contrib.SwitchMode.{Close, Open}
+import akka.stream.scaladsl.{Keep, Sink, Source}
+import akka.stream.testkit.scaladsl._
+import org.scalatest._
+import org.scalatest.Matchers._
+import org.scalatest.concurrent.ScalaFutures
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class ValveSpec extends WordSpec with ScalaFutures {
+
+  implicit val system = ActorSystem()
+  implicit val materializer = ActorMaterializer()
+  implicit val executionContext = materializer.executionContext
+
+  "A closed valve" should {
+
+    "emit only 3 elements into a sequence when the valve is switched to open" in {
+
+      val (switch, seq) = Source(1 to 3)
+        .viaMat(new Valve(SwitchMode.Close))(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+      after(100 millis, system.scheduler) {
+        val future: Future[Boolean] = switch.flip(Open)
+        whenReady(future, timeout(200 millis)) {
+          _ shouldBe true
+        }
+
+        future
+      }
+
+      whenReady(seq, timeout(200 millis)) {
+        _ should contain inOrder(1, 2, 3)
+      }
+    }
+
+    "emit only 5 elements when the valve is switched to open" in {
+      val (switch, probe) = Source(1 to 5)
+        .viaMat(new Valve(SwitchMode.Close))(Keep.right)
+        .toMat(TestSink.probe[Int])(Keep.both)
+        .run()
+
+      probe.request(2)
+      probe.expectNoMsg(100 millis)
+
+      whenReady(switch.flip(Open)) {
+        _ shouldBe true
+      }
+
+      probe.expectNext shouldBe 1
+      probe.expectNext shouldBe 2
+
+      probe.request(3)
+      probe.expectNext shouldBe 3
+      probe.expectNext shouldBe 4
+      probe.expectNext shouldBe 5
+
+      probe.expectComplete()
+    }
+
+
+    "emit only 3 elements when the valve is switch to open/close/open" in {
+      val ((sourceProbe, switch), sinkProbe) = TestSource.probe[Int]
+        .viaMat(Valve())(Keep.both)
+        .toMat(TestSink.probe[Int])(Keep.both)
+        .run()
+
+
+      sinkProbe.request(1)
+      whenReady(switch.flip(Close)) {
+        _ shouldBe true
+      }
+      sourceProbe.sendNext(1)
+      sinkProbe.expectNoMsg(100 millis)
+
+      whenReady(switch.flip(Open)) {
+        _ shouldBe true
+      }
+      sinkProbe.expectNext shouldEqual 1
+
+      whenReady(switch.flip(Close)) {
+        _ shouldBe true
+      }
+      whenReady(switch.flip(Open)) {
+        _ shouldBe true
+      }
+      sinkProbe.expectNoMsg(100 millis)
+
+      sinkProbe.request(1)
+      sinkProbe.request(1)
+      sourceProbe.sendNext(2)
+      sourceProbe.sendNext(3)
+      sourceProbe.sendComplete()
+
+      sinkProbe.expectNext shouldBe 2
+      sinkProbe.expectNext shouldBe 3
+
+      sinkProbe.expectComplete()
+    }
+
+    "return false when the valve is already closed" in {
+      val (switch, probe) = Source(1 to 5)
+        .viaMat(Valve(SwitchMode.Close))(Keep.right)
+        .toMat(TestSink.probe[Int])(Keep.both)
+        .run()
+
+      whenReady(switch.flip(Close)) { element =>
+        element should be(false)
+      }
+      whenReady(switch.flip(Close)) { element =>
+        element should be(false)
+      }
+    }
+
+    "emit nothing when the source is empty" in {
+      val (switch, seq) = Source.empty
+        .viaMat(Valve(SwitchMode.Close))(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+
+      whenReady(seq, timeout(200 millis)) {
+        _ shouldBe empty
+      }
+    }
+
+    "emit nothing when the source is failing" in {
+      val (switch, seq) = Source.failed(new IllegalArgumentException("Fake exception"))
+        .viaMat(Valve(SwitchMode.Close))(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+      whenReady(seq.failed) { e =>
+        e shouldBe an[IllegalArgumentException]
+      }
+    }
+
+  }
+
+  "A opened valve" should {
+
+    "emit 5 elements after it has been close/open" in {
+      val (switch, probe) = Source(1 to 5)
+        .viaMat(Valve())(Keep.right)
+        .toMat(TestSink.probe[Int])(Keep.both)
+        .run()
+
+      probe.request(2)
+      probe.expectNext() shouldBe 1
+      probe.expectNext() shouldBe 2
+
+      whenReady(switch.flip(Close)) {
+        _ shouldBe true
+      }
+
+      probe.request(1)
+      probe.expectNoMsg(100 millis)
+
+      whenReady(switch.flip(Open)) {
+        _ shouldBe true
+      }
+      probe.expectNext() shouldBe 3
+
+      probe.request(2)
+      probe.expectNext() shouldBe 4
+      probe.expectNext() shouldBe 5
+
+      probe.expectComplete()
+    }
+
+    "return false when the valve is already opened" in {
+      val (switch, probe) = Source(1 to 5)
+        .viaMat(Valve())(Keep.right)
+        .toMat(TestSink.probe[Int])(Keep.both)
+        .run()
+
+      whenReady(switch.flip(Open)) {
+        _ shouldBe false
+      }
+      whenReady(switch.flip(Open)) {
+        _ shouldBe false
+      }
+    }
+
+    "emit only 3 elements into a sequence" in {
+
+      val (switch, seq) = Source(1 to 3)
+        .viaMat(Valve())(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+      whenReady(seq, timeout(200 millis)) {
+        _ should contain inOrder(1, 2, 3)
+      }
+    }
+
+    "emit nothing when the source is empty" in {
+      val (switch, seq) = Source.empty
+        .viaMat(Valve())(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+
+      whenReady(seq, timeout(200 millis)) {
+        _ shouldBe empty
+      }
+
+    }
+
+    "emit nothing when the source is failing" in {
+      val (switch, seq) = Source.failed(new IllegalArgumentException("Fake exception"))
+        .viaMat(Valve())(Keep.right)
+        .toMat(Sink.seq)(Keep.both)
+        .run()
+
+      whenReady(seq.failed) { e =>
+        e shouldBe an[IllegalArgumentException]
+      }
+    }
+  }
+}


typesafehub-validator:
Hi @regis-leray,

ktoso:
OK TO TEST

ktoso:
I found myself implementing such thing on and off sometimes, good idea to include here i think.. Will review soon, thanks !

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,62 @@
...
+    import shape._
+
+    var bufferedElement = Option.empty[A]
+
+    val switch = new ValveSwitch {
+      val callback = getAsyncCallback[A](push(out, _))
+
+      override def open(): Unit = {
+        mode = ValveMode.Open
+        bufferedElement.foreach(callback.invoke)
```
regis-leray: Can you give more details about this race condition who could happen ? 
Have you already a suggestion to implements the trigger ? how should i properly handle this use case ?

Are you proposing to remove the open() and close() from the switch and to replace by a more generic method like toggle() ?

regis-leray:
```

Comment (regis-leray) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: I saw a comment concerning the race condition 
```unfortunately you're still doing unsafe things (you have a race-condition in your accesses to bufferedElement)``` thats make totally sense

Comment (viktorklang) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: You need to put the logic inside the callback instead of around the invocation of the callback.

**Comment by regis-leray** [comment]
@viktorklang thanks for your comments.
The code is much safer, but the only thing now it is less obvious in the code what it the intent when we call the switch: 


```scala
 val (switch, seq) = Source(1 to 3)
        .viaMat(new Valve(ValveMode.Closed))(Keep.right)
        .toMat(Sink.seq)(Keep.both)
        .run()

switch.toggle() 
switch.toggle()
switch.toggle()

// at the end the switch is Open or Close ??
```

viktorklang:
Have toggle() return the new state?

-- 
Cheers,
√

viktorklang:
Or have it called flip and have it take the expected new state and return a
boolean if the flip was successful?

-- 
Cheers,
√

regis-leray:
@viktorklang  ok i see i like the idea of

```
trait ValveSwitch {
  def flip(mode: ValveMode): Future[Boolean]
}

switch.flip(ValveMode.Open) // it is more explicit
```

To retreive the status i will have to change the return to a future since the operation is asynchronous

regis-leray:
@viktorklang  i implemented the change.

regis-leray:
@viktorklang @ktoso no more comments ?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,100 @@
...
+            if (mode == SwitchMode.Close) {
+              mode = SwitchMode.Open
+              element.foreach(push(out, _))
+            }
+            else {
+              mode = SwitchMode.Close
+            }
+          } match {
+            case Success(_) => promise.success(currentMode != _mode)
+            case Failure(ex) => promise.failure(ex)
```
regis-leray: the only case i can see where an exception can happen is in the ```push()``` logic (but i wasn t sure that could be happen).
Maybe the Try logic is a little bit too much

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+      * Buffer the element to push if the valve was close if a push was requested
+      */
+    var bufferedElement = Option.empty[A]
+
+    /**
+      * Keep track of the pull request when the valve is close
+      */
+    var pullInRequested = Option.empty[Unit]
```
viktorklang: @ktoso isn't it possible to query the GraphStage whether a pull has been requested? (rather than having to add a tracking-var?)

johanandren: `hasBeenPulled(in)` ?

viktorklang: No, whether a pull has come in from downstream but hasn't been requested to upstream yet.

johanandren: Ah, I think that would be `isAvailable(out)` then, because downstream will only be available after it has requested an element but the stage didn't push.

ktoso: How would you track that in a single call? Remember we're not a processor, we're an octopus (many ports possible) 🐙  
You can check specific upstream port and specific downstream port states and then decide.

johanandren: But, isn't this a simple `FlowShape`?

ktoso: Nevermind, I see you only meant "the other direction". isAvailable sounds like the way to go as Johan said.

viktorklang: @ktoso `outputs.fold(false)(_ || isAvailable(_))` ;-)

ktoso: > But, isn't this a simple FlowShape?

Just saying that GraphStage does not have a concept of "the upstream" and "the downstream" upon which we could say "the downstream did not cause a pull yet", since it's independent of shape

viktorklang: @ktoso Technically correct. This though is a FlowShape :)

viktorklang: @regis-leray Yes. But you won't need to do that since FlowShape only has a single `out`. So `isAvailable(out)` should be good enough

regis-leray: ok thanks i did the change

regis-leray:
@viktorklang any other comments ?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+
+    override def onPull(): Unit = {
+      if(isOpen) {
+        pull(in) //request the next element on in port
+      }
+    }
+
+    private def isClose = mode == SwitchMode.Close
```
johanandren: `isClosed`

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+      */
+    var bufferedElement = Option.empty[A]
+
+    val switch = new ValveSwitch {
+
+      override def flip(_mode: SwitchMode): Future[Boolean] = {
+        val promise = Promise[Boolean]()
+
+        val callback = getAsyncCallback[Option[A]]{ element =>
+          val currentMode = mode
```
johanandren: indentation is off from next line and further

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+            if (isClose) {
+              element.foreach(push(out, _))
+              bufferedElement = Option.empty
+
+              if (isAvailable(out)){ // if a pull in has been requested we trigger pull(In)
+                pull(in)
+              }
+
+              mode = SwitchMode.Open
+            }
```
johanandren: weird linefeed, don't we have auto-formatting in the sbt build fixing things like this when you compile?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  override val shape = FlowShape(Inlet[A]("valve.in"), Outlet[A]("valve.out"))
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, ValveSwitch) = {
+    val logic = new ValveGraphStageLogic(shape, mode)
+    (logic, logic.switch)
+  }
+
+  private class ValveGraphStageLogic(shape: Shape, var mode: SwitchMode) extends GraphStageLogic(shape) with InHandler with OutHandler {
+    import shape._
```
johanandren: what is this import for?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+
+        callback.invoke(bufferedElement)
+        promise.future
+      }
+    }
+
+    setHandlers(in, out, this)
+
+    override def onPush(): Unit = {
+      val element = grab(in) //acquires the element that has been received during an onPush
```
johanandren: this comment can go, it explains how the GraphStageLogic works rather than anything specific about this stage, same for the one below, if you want to add an explanation I think the interesting part to comment would be when buffering, something like "If the valve was closed after a pull was done upstream but before the element arrived it may arrive when the valve is closed and must be buffered until the valve is opened again".

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+    val switch = new ValveSwitch {
+
+      override def flip(_mode: SwitchMode): Future[Boolean] = {
+        val promise = Promise[Boolean]()
+
+        val callback = getAsyncCallback[Option[A]]{ element =>
+          val currentMode = mode
+            if (isClose) {
+              element.foreach(push(out, _))
+              bufferedElement = Option.empty
```
johanandren: Minor style detail: using `None` directly is the usual way to do this (And `empty` for the various collections types)

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+/**
+  * Pause/ Resume a Flow
+  */
+trait ValveSwitch {
+  /**
+    * Change the state of the valve
+    *
+    * @param mode expected mode to switch on
+    *
+    * @return
```
johanandren: "A future that completes with `true` if the mode did change and `false` if it already was in the requested mode"

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+    *         Future[true] if the mode change
+    *         Future[false] if the mode didn't change
+    */
+  def flip(mode: SwitchMode): Future[Boolean]
+}
+
+/**
+  * Valve permit to control a flow (pause/resume) by using a ValveSwitch.
```
johanandren: Please indent scaladoc like javadoc (single space before star here).

Also, I think this entire doc could be condensed into something more clear, for example: "Materializes into a [[ValeSwitch]] which provides a the method `flip` that stops or or restarts the flow of elements passing through the stage. As long as the valve is closed it will backpressure. 

Note that closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost."

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+              if (isAvailable(out)){ // if a pull in has been requested we trigger pull(In)
+                pull(in)
+              }
+
+              mode = SwitchMode.Open
+            }
+            else {
+              mode = SwitchMode.Close
+            }
+           promise.success(currentMode != _mode)
```
viktorklang: You're switching modes even if currentMode == _mode, that seems like a bug.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+  * After calling [[ValveSwitch#flip(mode: SwitchMode)]] the running instance of the [[Graph]] of [[FlowShape]] that materialized to the
+  * [[Valve]] will change the state of the valve
+  *
+  * After calling [[ValveSwitch#flip(Close)]]: the valve state is changed to Close if a push is requested the current element is saved in a buffer and if a pull is trigger the called will be saved.
+  *
+  * After calling [[ValveSwitch#flip(Open)]]: the valve state is changed to Open if a buffered Element is detected it will be automatically be push and a pull will be trigger if one was requested
+  *
+  * @param mode : state of the valve at the startup of the flow (by default Open)
+  */
+class Valve[A](mode: SwitchMode = SwitchMode.Open) extends GraphStageWithMaterializedValue[FlowShape[A, A], ValveSwitch] {
```
johanandren: I think factory methods for creating the Scala DSL and Java DSL flows would be nice. Look at the other existing stages here in contrib for examples.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+
+              mode = SwitchMode.Open
+            }
+            else {
+              mode = SwitchMode.Close
+            }
+           promise.success(currentMode != _mode)
+        }
+
+        callback.invoke(bufferedElement)
```
viktorklang: Why are you accessing `bufferedElement` outside of the callback?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
johanandren: I don't think it does any damage though, because it's not happening concurrently, so buffer will be drained/in not available after the first change to the same state.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    setHandlers(in, out, this)
+
+    override def onPush(): Unit = {
+      val element = grab(in) //acquires the element that has been received during an onPush
+      if (isOpen) {
+        push(out, element) //push directly the element on the out port
+      } else {
+        bufferedElement = Some(element)
```
johanandren: It never is because it can only get a value when close was called after pull, after that there will be no push until the valve is opened again

contrib/src/main/scala/akka/stream/contrib/Valve.scala
johanandren: Oh, I missed that, that is not allowed!

contrib/src/main/scala/akka/stream/contrib/Valve.scala
johanandren: Ah, I failed to notice. Please define `in` and `out` as fields on your GraphStage like in the samples in the docs and that way you will have them in scope already. (see samples here for example: http://doc.akka.io/docs/akka/2.4.11/scala/stream/stream-customize.html#Custom_processing_with_GraphStage)

regis-leray:
@johanandren @viktorklang i addressed all your comments. Thanks

contrib/src/main/scala/akka/stream/contrib/Valve.scala
regis-leray: you are right victor it was not properly handle.

The current situation was not working properly as you suggested !

```
 val (switch, probe) = Source(1 to 5)
        .viaMat(Valve(SwitchMode.Close))(Keep.right)
        .toMat(TestSink.probe[Int])(Keep.both)
        .run()

switch.flip(Close) // should return false
switch.flip(Close) // should return false
```

I had a unit test to cover this but the assertion were done with ```==``` and not with ```should be()```

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+  * Materializes into a [[ValveSwitch]] which provides a the method flip that stops or restarts the flow of elements passing through the stage. As long as the valve is closed it will backpressure.
+  *
+  * Note that closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost.
+  *
+  * @param mode : state of the valve at the startup of the flow (by default Open)
+  */
+class Valve[A](mode: SwitchMode) extends GraphStageWithMaterializedValue[FlowShape[A, A], ValveSwitch] {
+
+  private val in: Inlet[A] = Inlet[A]("valve.in")
+  private val out: Outlet[A] = Outlet[A]("valve.out")
```
johanandren: Don't make them private, being public allows them to be used for wiring with other stages in the GraphDSL

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+    * @return [[Valve]] Opened instance
+    *
+    */
+  def apply[A](): Valve[A] = new Valve[A](SwitchMode.Open)
+
+  /**
+    * Factory for [[Valve]] instances.
+    *
+    * @tparam A input and output type
+    * @return [[Valve]] instance
```
johanandren: I think these two (@param and @tparam) are superflous, that it returns an instance of `Valve` is obvious from the type signature. 

The factory methods could be made more self-documenting by not typing with the concrete type but with `Graph[FlowShape[A, A], ValveSwitch]` - that way there will not really be a need for any docs except the single line already present.

Also, again the scaladoc format should be a single space indent, like javadocs, so:
```
/**
 * 
 * 
```
and not as it is now:
```
/** 
  *
  *
```

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler}
+import akka.stream._
+import akka.stream.contrib.SwitchMode.{Close, Open}
+
+import scala.concurrent.{Future, Promise}
+
+/**
+  * Pause/ Resume a Flow
+  */
+trait ValveSwitch {
```
johanandren: Can be made sealed as we do not expect any other implementations than our private one

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+  def create[A](mode: SwitchMode): Valve[A] = new Valve[A](mode)
+
+}
+
+/**
+  * Materializes into a [[ValveSwitch]] which provides a the method flip that stops or restarts the flow of elements passing through the stage. As long as the valve is closed it will backpressure.
+  *
+  * Note that closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost.
+  *
+  * @param mode : state of the valve at the startup of the flow (by default Open)
```
johanandren: remove colon

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, ValveSwitch) = {
+    val logic = new ValveGraphStageLogic(shape, mode)
+    (logic, logic.switch)
+  }
+
+  private class ValveGraphStageLogic(shape: Shape, var mode: SwitchMode) extends GraphStageLogic(shape) with InHandler with OutHandler {
+    /**
```
johanandren: just a comment, not scaladoc, also, it's just a single line so `//` will do

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+    */
+  def create[A](): Valve[A] = new Valve[A](SwitchMode.Open)
+
+  /**
+    * Java API: Factory for [[Valve]] instances.
+    *
+    * @tparam A input and output type
+    * @return [[Valve]] instance
+    */
+  def create[A](mode: SwitchMode): Valve[A] = new Valve[A](mode)
```
viktorklang: apply(mode)

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/**
+  * Materializes into a [[ValveSwitch]] which provides a the method flip that stops or restarts the flow of elements passing through the stage. As long as the valve is closed it will backpressure.
+  *
+  * Note that closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost.
+  *
+  * @param mode : state of the valve at the startup of the flow (by default Open)
+  */
+class Valve[A](mode: SwitchMode) extends GraphStageWithMaterializedValue[FlowShape[A, A], ValveSwitch] {
```
viktorklang: not `final`?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+              }
+
+              mode = SwitchMode.Open
+              true
+          }
+
+          promise.success(succeed)
+        }
+
+        callback.invoke(NotUsed)
```
viktorklang: You should have the Mode be the parameter of the async callback, then you would onyl have to allocate a single callback instead of one per flip-invocation.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+
+    override def onPull(): Unit = {
+      if (isOpened) {
+        pull(in)
+      }
+    }
+
+    private def isOpened = mode == SwitchMode.Open
```
viktorklang: isOpen?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
johanandren: I don't think toggling the mode so often that allocations/gc becomes a performance problem, but having a re-usable callback would be more idiomatic than creating one for each toggle.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,150 @@
...
+    */
+  def apply[A](mode: SwitchMode): Valve[A] = new Valve[A](mode)
+
+  /**
+    * Java API: Factory for [[Valve]] instances.
+    *
+    * @tparam A input and output type
+    * @return [[Valve]] Opened instance
+    */
+  def create[A](): Valve[A] = new Valve[A](SwitchMode.Open)
```
johanandren: Never mind that last comment, I was thinking the comment was somewhere else...

contrib/src/main/scala/akka/stream/contrib/Valve.scala
regis-leray: I try what you suggested and as i expected i got in my unit test:
```assertion failed: received unexpected message OnError(java.lang.IllegalStateException: Promise already completed.```

File: contrib/src/main/scala/akka/stream/contrib/Valve.scala (hunk 21 above)
Comment (regis-leray) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: ? I dont get your comment it is for the ***Java API***. as suggested @johanandren  i provided a factory to instanciate the Valve.

File: contrib/src/main/scala/akka/stream/contrib/Valve.scala (hunk 25 above)
Comment (regis-leray) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: 👍

File: contrib/src/main/scala/akka/stream/contrib/Valve.scala (hunk 17 above)
Comment (regis-leray) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: My intellij doesnt seems to pickup the format configuration, im looking into it

File: contrib/src/main/scala/akka/stream/contrib/Valve.scala (hunk 23 above)
Comment (regis-leray) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: @johanandren yeah good idea

**Comment by regis-leray** [comment]
All comments addressed, thanks for your great comment i learn a lot :)

**Comment by viktorklang** [comment]
@regis-leray Thanks for being awesome!

:+1:

**Comment by regis-leray** [comment]
@2m @ktoso  @johanandren any change we can merge this ? and do a release ?

File: contrib/src/main/scala/akka/stream/contrib/Valve.scala [hunk 26]
```diff
@@ -0,0 +1,138 @@
...
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, ValveSwitch) = {
+    val logic = new ValveGraphStageLogic(shape, mode)
+    (logic, logic.switch)
+  }
+
+  private class ValveGraphStageLogic(shape: Shape, var mode: SwitchMode) extends GraphStageLogic(shape) with InHandler with OutHandler {
+    val switch = new ValveSwitch {
+
+      val callback = getAsyncCallback[(SwitchMode, Promise[Boolean])] {
```
drewhk: I was a bit surprised why an asyncCallback is needed instead of a dumb AtomicBoolean but I got it now: You might need to kick out the stage into execution when it has an element buffered. I guess this is the simplest thing that can work, but I am a bit unhappy with the Future[Boolean] return value.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,138 @@
...
+    var bufferedElement = Option.empty[A]
+
+    setHandlers(in, out, this)
+
+    override def onPush(): Unit = {
+      val element = grab(in)
+      if (isOpen) {
+        push(out, element)
+      } else {
+        bufferedElement = Some(element)
```
drewhk: no need for the buffer variable. Just don't grab() the element if you don't need it (and check for the presence of an element with `isAvailable`). See http://doc.akka.io/docs/akka/2.4/scala/stream/stream-customize.html#Port_states__InHandler_and_OutHandler for the state diagram for input ports.

regis-leray: thanks great idea. I implemented as you suggested.

contrib/src/main/scala/akka/stream/contrib/Valve.scala
drewhk: You cannot reliably return a Boolean due to the fact that an AsyncCallback is needed to wake up the stage and hence the state transition is async and can be in an undecided race with another call which is not yet won by either parties at the point of call. My comment was an observation of this :)

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -0,0 +1,140 @@
...
+        case (flipToMode, promise) =>
+          val succeed = mode match {
+            case _ if flipToMode == mode => false
+
+            case Open =>
+              mode = SwitchMode.Close
+              true
+
+            case Close =>
+              grabIfAvailable().foreach(push(out, _))
```
drewhk: Isn't it simpler to just write

```scala
if (isAvailable(in)) push(out, grab(in))
else if (isAvailable(out)) pull(in)
```

And then no need to allocate an Option in the end

regis-leray: done !

samidalouche: @drewhk I am not convinced of this change. I am seeing a problem in a real-life use-case that ends up raising the following exception

```java.lang.IllegalArgumentException: requirement failed: Cannot pull port (valve.in) twice```

I am trying to reproduce the exact scenario in a unit-test (only got an end-to-end test for my system to fail), but the scenario basically involves having an open valve, closing it, then re-opening it.

Comment (samidalouche) [inline contrib/src/main/scala/akka/stream/contrib/Valve.scala]: @drewhk @regis-leray I got to reproduce the issue with a simple test

```scala
  "An open valve" should {
    "not pull elements again when closed and re-opened" in {

      val (switch, done) = Source.maybe[Int]
        .viaMat(new Valve(SwitchMode.Open))(Keep.right)
        .toMat(Sink.ignore)(Keep.both)
        .run()

      val ok = for {
        _ <- switch.flip(SwitchMode.Close)
        _ <- switch.flip(SwitchMode.Open)
        _ <- done
      } yield true

      whenReady(ok, timeout = Timeout(5 seconds)) {
        _ shouldBe true
      }
    }
  }
```

samidalouche: Reported a bug (https://github.com/akka/akka-stream-contrib/issues/84) to track this issue separately

PR #76, opened 2016-10-26
add cluster http api #21172

hveiga:
Initial development for a cluster http api. Ticket: https://github.com/akka/akka/issues/21172

### DIFF
--- a/cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
@@ -1,0 +1,145 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.cluster.http
+
+import akka.actor.AddressFromURIString
+import akka.cluster.{Cluster, Member}
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.RouteResult
+import akka.stream.ActorMaterializer
+import spray.json._
+
+import scala.concurrent.Future
+
+final case class ClusterUnreachableMember(node: String, observedBy: Seq[String])
+final case class ClusterMember(node: String, nodeUid: String, status: String, roles: Set[String])
+final case class ClusterMembers(selfNode: String, members: Set[ClusterMember], unreachable: Seq[ClusterUnreachableMember])
+final case class ClusterHttpApiMessage(message: String)
+
+object ClusterHttpApiOperation extends Enumeration {
+  val DOWN, LEAVE, JOIN = Value
+
+  def find(name: String) = try {
+    Some(withName(name))
+  } catch {
+    case e: NoSuchElementException =>
+      None
+  }
+}
+
+trait ClusterHttpApiJsonProtocol extends SprayJsonSupport with DefaultJsonProtocol {
+  implicit val clusterUnreachableMemberFormat = jsonFormat2(ClusterUnreachableMember)
+  implicit val clusterMemberFormat = jsonFormat4(ClusterMember)
+  implicit val clusterMembersFormat = jsonFormat3(ClusterMembers)
+  implicit val clusterMemberMessageFormat = jsonFormat1(ClusterHttpApiMessage)
+}
+
+trait ClusterHttpApiHelper extends ClusterHttpApiJsonProtocol {
+  def memberToClusterMember(m: Member): ClusterMember = {
+    ClusterMember(s"${m.uniqueAddress.address}", s"${m.uniqueAddress.uid}", s"${m.status}", m.roles)
+  }
+}
+
+object ClusterHttpRoute extends ClusterHttpApiHelper {
+
+  private def routeGetMembers(cluster: Cluster) = {
+    get {
+      complete {
+        val members = cluster.readView.state.members.map(memberToClusterMember)
+
+        val unreachable = cluster.readView.reachability.observersGroupedByUnreachable.toSeq.sortBy(_._1).map {
+          case (subject, observers) ⇒
+            ClusterUnreachableMember(s"${subject.address}", observers.toSeq.sorted.map(m ⇒ s"${m.address}"))
+        }
+
+        ClusterMembers(s"${cluster.readView.selfAddress}", members, unreachable)
+      }
+    }
+  }
+
+  private def routePostMembers(cluster: Cluster) = {
+    post {
+      formField('address) { address ⇒
+        complete {
+          cluster.join(AddressFromURIString(address))
+          ClusterHttpApiMessage("Operation executed")
+        }
+      }
+    }
+  }
+
+  private def routeGetMember(cluster: Cluster, member: String) =
+    get {
+      complete {
+        cluster.readView.members.find(m ⇒ s"${m.uniqueAddress.address}" == member).map(memberToClusterMember)
+      }
+    }
+
+  private def routeDeleteMember(cluster: Cluster, member: String) =
+    delete {
+      cluster.readView.members.find(m ⇒ s"${m.uniqueAddress.address}" == member) match {
+        case Some(m) =>
+          cluster.leave(m.uniqueAddress.address)
+          complete(ClusterHttpApiMessage("Operation executed"))
+        case None =>
+          complete(StatusCodes.NotFound -> ClusterHttpApiMessage("Member not found"))
+      }
+    }
+
+  private def routePutMember(cluster: Cluster, member: String) =
+    put {
+      formField('operation) { operation =>
+        cluster.readView.members.find(m ⇒ s"${m.uniqueAddress.address}" == member) match {
+          case Some(m) =>
+            ClusterHttpApiOperation.find(operation.toUpperCase) match {
+              case Some(ClusterHttpApiOperation.DOWN) =>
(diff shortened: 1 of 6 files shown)


cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+import spray.json._
+
+import scala.concurrent.Future
+
+final case class ClusterUnreachableMember(node: String, observedBy: Seq[String])
+final case class ClusterMember(node: String, nodeUid: String, status: String, roles: Set[String])
+final case class ClusterMembers(selfNode: String, members: Set[ClusterMember], unreachable: Seq[ClusterUnreachableMember])
+final case class ClusterHttpApiMessage(message: String)
+
+object ClusterHttpApiOperation extends Enumeration {
```
drewhk: is there any benefit of Enumeration over having a trait and case objects extending it?

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+          ClusterHttpApiMessage("Operation executed")
+        }
+      }
+    }
+  }
+
+  private def routeGetMember(cluster: Cluster, member: String) =
+    get {
+      complete {
+        cluster.readView.members.find(m ⇒ s"${m.uniqueAddress.address}" == member).map(memberToClusterMember)
```
drewhk: generate the string for comparison upfront and store it in a val (instead of creating a new one for each invocation)

drewhk: (fix it at other places, too)

cluster-http/src/test/scala/akka/cluster/http/ClusterHttpApiSpec.scala
```diff
@@ -0,0 +1,256 @@
...
+
+        val clusterMember1 = Member(uniqueAddress1, Set())
+        val clusterMember2 = Member(uniqueAddress2, Set())
+        val currentClusterState = CurrentClusterState(SortedSet(clusterMember1, clusterMember2))
+
+        val unreachable = Map(
+          UniqueAddress(address3, 2) -> Set(uniqueAddress1, uniqueAddress2)
+        )
+
+        val mockedCluster = mock(classOf[Cluster])
```
drewhk: not big fan of mocking, but this might actually simplify the test and library here (compared to a full-blown multi-jvm test).

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
ktoso: +1 for sealed trait and case objects extending it

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+  implicit val clusterMemberMessageFormat = jsonFormat1(ClusterHttpApiMessage)
+}
+
+trait ClusterHttpApiHelper extends ClusterHttpApiJsonProtocol {
+  def memberToClusterMember(m: Member): ClusterMember = {
+    ClusterMember(s"${m.uniqueAddress.address}", s"${m.uniqueAddress.uid}", s"${m.status}", m.roles)
+  }
+}
+
+object ClusterHttpRoute extends ClusterHttpApiHelper {
```
ktoso: Route or Routes, do we have an opinion for consistent naming here @jrudolph ?
I normally always call classes `Routes` but we don't really have a recommendation in Akka/Spray itself.

ktoso: Let's add a small documentation (what end point and what it does here), so people can figure that out even if they don't have sources of this class (and no reference documentation).

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  private def routePutMember(cluster: Cluster, member: String) =
+    put {
+      formField('operation) { operation =>
+        cluster.readView.members.find(m ⇒ s"${m.uniqueAddress.address}" == member) match {
+          case Some(m) =>
+            ClusterHttpApiOperation.find(operation.toUpperCase) match {
+              case Some(ClusterHttpApiOperation.DOWN) =>
+                cluster.down(m.uniqueAddress.address)
+                complete(ClusterHttpApiMessage("Operation executed"))
```
ktoso: Same here, let's be explicit "Downing node ..."

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+              case Some(ClusterHttpApiOperation.LEAVE) =>
+                cluster.leave(m.uniqueAddress.address)
+                complete(ClusterHttpApiMessage("Operation executed"))
+              case _ =>
+                complete(StatusCodes.NotFound -> ClusterHttpApiMessage("Operation not supported"))
+            }
+          case None =>
+            complete(StatusCodes.NotFound -> ClusterHttpApiMessage("Member not found"))
```
ktoso: good 👍

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+          case Some(m) =>
+            ClusterHttpApiOperation.find(operation.toUpperCase) match {
+              case Some(ClusterHttpApiOperation.DOWN) =>
+                cluster.down(m.uniqueAddress.address)
+                complete(ClusterHttpApiMessage("Operation executed"))
+              case Some(ClusterHttpApiOperation.LEAVE) =>
+                cluster.leave(m.uniqueAddress.address)
+                complete(ClusterHttpApiMessage("Operation executed"))
+              case _ =>
+                complete(StatusCodes.NotFound -> ClusterHttpApiMessage("Operation not supported"))
```
ktoso: Let's use `400 Bad Request` instead here, it's a client error here, not a not found.

cluster-http/src/main/scala/akka/cluster/http/ClusterHttpApi.scala
```diff
@@ -0,0 +1,145 @@
...
+            }
+          case None =>
+            complete(StatusCodes.NotFound -> ClusterHttpApiMessage("Member not found"))
+        }
+      }
+    }
+
+  def apply(cluster: Cluster) =
+    pathPrefix("members") {
```
ktoso: I wonder if the path should be configurable by users hm... maybe add overloaded apply with 2 parameters?

ktoso:
Very cool feature, we've been long waiting to get this one :-)
Impl also good, just some minor nitpicks, thanks!

I think this would perhaps better to put into http://github.com/akka/akka-http thus it'd be always up to date and not need to link to akka-http as external dependency. Happy to keep releasing it along with Akka HTTP. Suggest to move it over there and add a small documentation page for it? We need to make sure it's usable from Java as well.

Again, great feature - thanks!

patriknw:
It's difficult to find a good home for this. This was my thinking:

* can't be in akka-cluster or any akka/akka module, since it has dependency to akka-http (would be circular between the projects)
* can't be in akka-http because we don't want additional dependency to akka-cluster there
* felt like overkill to create yet another repository for this

patriknw:
if we want it in akka-http the dependency to akka-cluster should be defined as `optional`

ktoso:
It can be in akka-http's repository, but as a separate project: "akka-http-cluster-api",
that project would depend on cluster. No-one else is forced to get the cluster dependency then :-)

ktoso:
Made it bin-compat again

patriknw:
ok, that sounds even better
not sure about the `-api` part of the name
if `akka-http-cluster` is not descriptive enough I suggest `akka-http-cluster-management`

ktoso:
Good name `-management`, 👍 
Would you be able to submit this PR to the Akka HTTP repo under such name @hveiga?
Thanks a lot!

iravid:
This is awesome. 

Is there interest in providing a querying interface for cluster sharding? We wrote something internal  to get a JSON object of `Map[ShardRegionAddress, Map[ShardId, EntityId]]` over HTTP.

patriknw:
That sounds great to add more.

hveiga:
Thanks everyone for the feedback. I can definitely implement those changes and create a new PR to akka-http. I am looking forward to see this merged :).

patriknw:
thanks @hveiga , great work

hveiga:
Hey again. I have almost everything ready for the new PR. I have two questions: 
 - Under which documentation page should I create some documentation for this new module? Networking? Akka HTTP?
- I would like to ensure the code is also usable with Java as well. How can I achieve it? The obvious would be to create a Java project and test it out. But, is there other way?

I am planning to finalize those two things this week and create the new PR.

patriknw:
Somewhere here https://github.com/akka/akka-http/tree/master/docs

and after release we should mention it in the Akka Cluster docs.

I think docs example code is an excellent way to verify that the api is usable from Java. Add sample code in https://github.com/akka/akka-http/tree/master/docs and include it in the docs. Note that Paradox can show Java/Scala tabs in same page if that makes sense for this documentation. http://developer.lightbend.com/docs/paradox/latest/features/snippet-inclusion.html

patriknw:
This will be located in https://github.com/akka/akka-cluster-management

PR #83, opened 2016-11-10
Make akka-stream-testkit dependency provided (closes #82)

### LINKED ISSUE #82: contrib sub-project depends on akka-stream-testkit in Compile configuration
It looks like this is necessary for `Coroner`, but nevertheless I think it's a no-go to pull test libraries into `Compile` configuration. Now every project depending on akka-stream-contrib, like akka-sse, has a `Compile` dependency on akka-stream-contrib.

hseeberger:
See #82.


patriknw:
thanks, I'll cut a new release

PR #88, opened 2017-03-15
A potentially useful akka-stream stage for time-window processing

marcin-rzeznicki:
Please refer to https://github.com/akka/akka/issues/22420 where motivation behind elements being committed was discussed. It was suggested there ( https://github.com/akka/akka/issues/22420#issuecomment-284156185 ) that a PR containing this work should go to contrib repository. At the cost of violating DRY, I am pasting relevant elements of the original issue here for easier reference:

> I couldn't find any existing stage to accomplish simple time-window processing where data is aggregated for predefined amount of time and the computed aggregate is emitted once per this time. While conflateWithSeed fits the "aggregation" part of processing very nicely, I couldn't find any element that could steer the back-pressure to pull the aggregate once in a while and push it downstream. delay is close, but since it needs at least one element input buffer, one ends up with situation where, conceptually: at t0 data is pulled from conflate stage and put into buffer to be emitted at t1, at t1 the next element is pulled (aggregate of ]t0, t1]) to be emitted at t2 and so on, which means unnecessary delay of at least one "window". groupedWithin might be used for this case, but it keeps elements in memory till time-window expires, which is unnecessary in my case, as all I care about is aggregated data which can be computed incrementally. Therefore I ended up writing a custom stage called Pulse which can be seen as backpressuring delay with no buffering.

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/Pulse.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/Pulse.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.stream.Attributes
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.stage._
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Signals demand only once every [[interval]] (''pulse'') and then back-pressures. Requested element is emitted downstream if there is demand.
+ *
+ * @param interval ''pulse'' period
+ * @param initiallyOpen if `true` - emits the first available element before ''pulsing''
+ * @tparam T type of element
+ */
+final class Pulse[T](val interval: FiniteDuration, val initiallyOpen: Boolean = false) extends SimpleLinearGraphStage[T] {
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with InHandler with OutHandler {
+
+      setHandlers(in, out, this)
+
+      override def preStart(): Unit = if (!initiallyOpen) startPulsing()
+      override def onPush(): Unit = if (isAvailable(out)) push(out, grab(in))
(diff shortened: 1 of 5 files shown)


typesafehub-validator:
Hi @marcin-rzeznicki,

marcin-rzeznicki:
Hey @2m - pingin'

contrib/src/main/scala/akka/stream/contrib/Pulse.scala
```diff
@@ -0,0 +1,48 @@
...
+ * @tparam T type of element
+ */
+final class Pulse[T](val interval: FiniteDuration, val initiallyOpen: Boolean = false) extends SimpleLinearGraphStage[T] {
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new TimerGraphStageLogic(shape) with InHandler with OutHandler {
+
+      setHandlers(in, out, this)
+
+      override def preStart(): Unit = if (!initiallyOpen) startPulsing()
+      override def onPush(): Unit = if (isAvailable(out)) push(out, grab(in))
```
2m: There is no other `push(...)` in this stage. So if there is no demand from the downstream but elements keep coming from the upstream, elements will be lost here. I have added a failing test-case to this branch to illustrate this.

marcin-rzeznicki: Thanks, nice catch. So I guess I need to check `isAvailable(out)` on timer release

marcin-rzeznicki: Fixed in 6fa4d52 by checking if out signaled demand in timer

2m:
Thanks for a quick fix. PR validation failed in `TimeWindowSpecAutoFusingOn` this time. `sub.expectNext(timeWindow, 5)` can be too tight as that is the same as the interval of the Pulse stage. Let's double the wait time (`timeWindow * 2`) in the tests.

marcin-rzeznicki:
That's right - fixed in 4d9f67a . Thanks for the review!

2m:
Great, merged. :)

Issue #89, opened 2017-03-27
DelayFlow stage with flexible delay management


manonthegithub:
I needed flexible mechanism to make delay in stream under certain circumstances (for example if http server goes down not to oveload it with redundant requests, delay for some time after receiving some number of 500 error codes in a row) and thought that it can be useful to others. I'll submit PR with my implementation soon. It will be great, if you'll decide that it would be useful.

2m:
Fixed in #90

PR #90, opened 2017-03-27
DelayFlow with flexible DelayStrategy implementation #89

manonthegithub:
Implementation of flexible DelayStage with ability to create your own delay strategies with different delay policies. 

Refs #89

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/DelayFlow.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/DelayFlow.scala
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.contrib
+
+import akka.NotUsed
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.Attributes
+import akka.stream.contrib.DelayFlow.DelayStrategy
+import akka.stream.scaladsl.Flow
+import akka.stream.stage._
+import scala.concurrent.duration._
+
+object DelayFlow {
+
+  /**
+   * Flow with fixed delay for each element.
+   * @param fixedDelay value of the delay
+   */
+  def apply[T](fixedDelay: FiniteDuration): Flow[T, T, NotUsed] = DelayFlow[T](() => DelayStrategy.fixedDelay(fixedDelay))
+
+  /**
+   * Flow for universal delay management, allows to manage delay through [[DelayStrategy]].
+   * It determines delay for each ongoing element invoking `DelayStrategy.nextDelay(elem: T): FiniteDuration`.
+   * Implementing [[DelayStrategy]] with your own gives you flexible ability to manage delay value depending on coming elements.
+   * It is important notice that [[DelayStrategy]] can be stateful.
+   * There are also predefined strategies, see [[DelayStrategy]] companion object's methods.
+   *
+   *
+   * For example:
+   * {{{
+   * //delay, infinitely increasing by `1 second` on every Failure
+   *  new DelayStrategy[Try[AnyRef]]{
+   *    var delay = Duration.Zero
+   *    override def nextDelay(elem: Try[AnyRef]): FiniteDuration = {
+   *      if(elem.isFailure){
+   *        delay += (1 second)
+   *      }
+   *      delay
+   *    }
+   *  }
+   * }}}
+   * @param strategySupplier creates new [[DelayStrategy]] object for each materialization
+   * @see [[DelayStrategy]]
+   */
+  def apply[T](strategySupplier: () => DelayStrategy[_ >: T]): Flow[T, T, NotUsed] = Flow.fromGraph(new DelayFlow[T](strategySupplier))
+
+  object DelayStrategy {
+
+    /**
+     * Fixed delay strategy, always returns constant delay for any element.
+     * @param delay value of the delay
+     */
+    def fixedDelay(delay: FiniteDuration): DelayStrategy[Any] = new DelayStrategy[Any] {
+      override def nextDelay(elem: Any): FiniteDuration = delay
+    }
+
+    /**
+     * Strategy with linear increasing delay.
+     * It starts with `initialDelay` for each element,
+     * increases by `increaseStep` every time when `needsIncrease` returns `true` up to `maxDelay`,
+     * when `needsIncrease` returns `false` it resets to `initialDelay`.
+     * @param increaseStep step by which delay is increased
+     * @param needsIncrease if `true` delay increases, if `false` delay resets to `initialDelay`
+     * @param initialDelay initial delay for each of elements
+     * @param maxDelay limits maximum delay
+     */
+    def linearIncreasingDelay[T](
+      increaseStep:  FiniteDuration,
+      needsIncrease: T => Boolean,
+      initialDelay:  FiniteDuration = Duration.Zero,
+      maxDelay:      Duration       = Duration.Inf
+    ): DelayStrategy[T] = {
+      require(increaseStep > Duration.Zero, "Increase step must be positive")
+      require(maxDelay > initialDelay, "Max delay must be bigger than initial delay")
+
+      new DelayStrategy[T] {
+
+        private var delay = initialDelay
+
+        override def nextDelay(elem: T): FiniteDuration = {
+          if (needsIncrease(elem)) {
+            val next = delay + increaseStep
+            if (next < maxDelay) {
+              delay = next
+            } else {
+              delay = maxDelay.asInstanceOf[FiniteDuration]
+            }
+
+          } else {
+            delay = initialDelay
+          }
+          delay
+        }
+
+      }
+
+    }
+
+  }
+
+  /**
+   * Allows to manage delay and can be stateful to compute delay for any sequence of elements,
+   * all elements go through nextDelay() updating state and returning delay for each element
+   */
+  trait DelayStrategy[T] {
+
+    /**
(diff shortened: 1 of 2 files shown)


contrib/src/main/scala/akka/stream/contrib/DelayFlow.scala
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    override def onPull(): Unit = {
+      pull(in)
+    }
+
+    override def onTimer(timerKey: Any): Unit = {
+      push(out, delayedElem)
+      if (isClosed(in)) {
+        super.onUpstreamFinish()
```
patriknw: `completeStage()` is probably more correct

manonthegithub:
@patriknw made improvements according to your comments

manonthegithub:
rebased PR to resolve conflict

manonthegithub:
Hey guys. There was one fail in  that is not related to my PR: 
`[info] TimeWindow flow
[info] - should aggregate data for predefined amount of time *** FAILED *** (634 milliseconds)`.

Also there were fails in fixed delay test. The problem is that message comes to `expectNoMsg()` before its timer triggers. What is timer precision? How long should be the gap for test to pass for sure?

contrib/src/test/scala/akka/stream/contrib/DelayFlowSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+      val elems = Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 0)
+
+      val probe = Source(elems)
+        .via(DelayFlow[Int](fixedDelay))
+        .runWith(TestSink.probe)
+        .request(elems.size)
+
+      elems.foreach(probe
+        //the timer is not precise enough and fails occasionally if exact delay value is used
+        .expectNoMsg(fixedDelay - (10 millis))
```
manonthegithub: @patriknw @johanandren 
This test works on my machine, but failed on build server.
Can you guys please advice the size of the gap for test to pass for sure?

arpanchaudhury: You can run **validatePullRequest** task as mentioned in the contribution guidelines (if available for this project).

https://github.com/akka/akka/blob/master/CONTRIBUTING.md

johanandren: Hmm, that the ci-servers are slow should never cause a message to be emitted faster than the delay, right? Is it really the `expectNoMsg` failing?

manonthegithub: I guess scheduler in TimerGraphStageLogic is not precise enough... can it be the case?

johanandren: It's the actor system scheduler, which can trigger late, but it should never be early. The general thing with the CI-servers is rather that they are slow, so things may take more time than expected. I think shorter times may indicate a bug lurking.

manonthegithub: I'll try find out what's wrong, but I was unable to reproduce it despite some attempts today.

manonthegithub:
I guess it could be working wrong because of dilated, also made `request()` closer to `expectNoMsg`

johanandren:
Now the last run failed just because I gave a too early LGTM :)

johanandren:
Please look into that failure (I see there was two other tests that also failed, so likely a slowness/timing failure, but still good to see if the test can be made not fail)

manonthegithub:
Ok, I will investigate it more.

manonthegithub:
Hey guys, I've got interesting results. I've corrected the test little bit:
```scala
"work with fixed delay" in {
      for(_ <- 1 to 1000) {
        val fixedDelay = 1 second

        val elems = Vector(1, 2)

        val probe = Source(elems)
          .log("l1", e => println(s"Test in $e: ${System.currentTimeMillis()}"))
          .via(DelayFlow[Int](fixedDelay.dilated))
          .log("l2", e => println(s"Test out $e: ${System.currentTimeMillis()}"))
          .runWith(TestSink.probe)

        elems.foreach(probe
          .request(1)
          .expectNoMsg(fixedDelay)
          .expectNext(_))

        probe.expectComplete()
      }

    }
```
And got interesting results in console:
```
.......
Test out 2: 1495794585819
Test in 1: 1495794585835
Test out 1: 1495794586866
Test in 2: 1495794586867
Test out 2: 1495794587870
Test in 1: 1495794587884
Test out 1: 1495794588929

assertion failed: received unexpected message OnNext(1)
java.lang.AssertionError: assertion failed: received unexpected message OnNext(1)
	at scala.Predef$.assert(Predef.scala:219)
	at akka.testkit.TestKitBase.expectNoMsg_internal(TestKit.scala:616)
	at akka.testkit.TestKitBase.expectNoMsg(TestKit.scala:612)
	at akka.testkit.TestKitBase.expectNoMsg$(TestKit.scala:612)
	at akka.testkit.TestKit.expectNoMsg(TestKit.scala:814)
	at akka.stream.testkit.TestSubscriber$ManualProbe.expectNoMsg(StreamTestKit.scala:550)
	at akka.stream.contrib.DelayFlowSpec.$anonfun$$init$$7(DelayFlowSpec.scala:42)
	at akka.stream.contrib.DelayFlowSpec.$anonfun$$init$$7$adapted(DelayFlowSpec.scala:43)

```

How can you explain this? What's wrong with test? @patriknw @johanandren

johanandren:
The scheduler cannot guarantee exactly 1000 ms, it depends on a big amount of parameters (gc, jvm, threads, how the load on your machine is), 45ms delay sounds a bit much but the next element is delayed 3ms so that may be random. Do a benchmark of many more elements, say 10 000 and look at the average and the min/max. Also, don't use `System.currentTimeMillis` which usually have a higher granularity than 1ms, use `System.nanoTime` instead.

manonthegithub:
@johanandren Yeah, I will do more sample measurements. But I am about test failure, `expectNoMsg(fixedDelay)` is set 1 second, the actual delay time is 1045ms, so why the test fails on `expectNoMsg`?

manonthegithub:
Isn't it a bug in TestSubscriber Probe and what size of the gap to avoid inaccuracy should be?

manonthegithub:
The notable point is that, now I reproduced it and it failed on 1011ms despite the fact that maximun delay in a row was 1022ms.

johanandren:
And those you measured with `nanoTime`, not `currentTimeMillis`?

manonthegithub:
yes
image

on the bottom row test failed

manonthegithub:
I've got more samples, there were 705 successful passes. The summary is (diff time in nanos):
image

manonthegithub:
I've done more research.
I decided to write test without delay stage to test `expectNoMsg`:
```scala
"test" in {
      for(_ <- 1 to 10000) {
        val blockingDispatcher = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor())

        val elems = Vector(1,2)

        val probe = Source(elems)
          .log("l1", e => println(s"Test in $e: ${System.nanoTime()}"))
          .mapAsync(1)(e => Future{Thread.sleep(1010); e}(blockingDispatcher))//1007 fails
          .log("l2", e => println(s"Test out $e: ${System.nanoTime()}"))
          .runWith(TestSink.probe)

        elems.foreach(_ should be(probe
          .request(1)
          .expectNoMsg(1 second)
          .expectNext()))

        probe
          .request(1)
          .expectNextOrComplete(1)

      }
    }
```
If I set `Thread.sleep` to `1007` or less it fails almost immediately. Setting it to `1010` made it to work for few hours and then it failed too with:
```
assertion failed: received unexpected message OnNext(1)
java.lang.AssertionError: assertion failed: received unexpected message OnNext(1)
	at scala.Predef$.assert(Predef.scala:219)
	at akka.testkit.TestKitBase.expectNoMsg_internal(TestKit.scala:616)
```
The measurements show that failure was on `1010481768` nanos.
WDYT @johanandren ?

manonthegithub:
For now I completely removed expectNoMsg from tests and replaces it with System.nanoTime calculations

manonthegithub:
Guys, there are some heavy delays on build servers, that fail tests. I can increase wait time in my tests but there are some more tests failing on timeout that are not dependent on this PR. What do you think is better solution for this?

raboof:
@manonthegithub I agree this is getting problematic. While travis can be somewhat jittery, I think we should improve our tests so that they're more resilient.

Feel free to create issues for specific unreliable tests, we'll use the 'failure' label to track them. Fixing those should be a priority.

I'm happy to restart builds after we verified that there's issues for the failures, but it'd be really nice to at least have the `DelayFlow` tests stable before merging this PR...

manonthegithub:
@raboof
I agree, but my tests largely depend on timing. If everything gets slow and messages get delayed in an unpredictable way, it is problematic to fix them, the only thing I can do is to increase wait time then, which becomes much longer than expected... So these tests will have the only guarantee that elems get delayed by `DelayFlow` for 'at least' the specified delay but this delay can last longer one would expect.

manonthegithub:
After increased delay `DelayFlow` tests passed. Can you please retrigger it several times for sure?

manonthegithub:
squashed and rebased

manonthegithub:
I guess, It can be merged, because there are no tests in my PR that are failing (I also increased timeout in them). The failed ones are #109 and #104. I you will have have time, please review latest version).

manonthegithub:
Great. Thanks!

PR #92, opened 2017-04-13
Akka 2.5 support + custom timeout

kciesielski:
This change breaks Akka 2.4 compatibility. We need new `akka-stream-contrib` in order to run tests for `akka-stream-kafka` with Akka 2.5.0 support.


johanandren:
It seems the guard rails for graph stages in 2.5.0 makes some bugs in the stages surface which needs to be fixed.

This also means we may have to either maintain a 2.4 branch of stream contrib or just enforce 2.5, I'm not it is worth it keeping two versions like that.

(Spontaneously, wouldn't it be better for reactive-kafka to just have it's own copy of that test and not depend on stream-contrib?)

jrudolph:
Here's the compatible version we use in akka-http to cross-compile 2.4 and 2.5: https://github.com/akka/akka-http/blob/master/akka-http-core/src/test/scala/akka/stream/testkit/Utils.scala

kciesielski:
@johanandren Good idea, it's definitely better to copy the needed code, reactive-kafka doesn't depend on Valve. I'll just use Utils posted here by @jrudolph, thanks :)

PR #94, opened 2017-04-29
Upgrade to 2.5.0

cchantep:


cchantep:
Still an issue :

```
[info] ValveSpec:
[info] A closed valve
[info] - should emit only 3 elements into a sequence when the valve is switched to open (201 milliseconds)
[info] - should emit only 5 elements when the valve is switched to open (151 milliseconds)
[info] - should emit only 3 elements when the valve is switch to open/close/open *** FAILED *** (41 milliseconds)
[info]   java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor
[info]   at akka.stream.stage.GraphStageLogic.interpreter(GraphStage.scala:291)
[info]   at akka.stream.stage.GraphStageLogic$$anon$1.invoke(GraphStage.scala:956)
[info]   at akka.stream.contrib.Valve$ValveGraphStageLogic$$anon$1.flip(Valve.scala:100)
[info]   at akka.stream.contrib.ValveSpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(ValveSpec.scala:79)
[info]   at akka.stream.contrib.ValveSpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(ValveSpec.scala:72)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
```

2m:
Switching to the 2.5.0 (BTW 2.5.1 is now the latest) swaps in new materializer which is more likely to reveal timing issues because materialization is much faster.

The underlying issue is revealed here is https://github.com/akka/akka/issues/20503

A workaround here could be to expose a `Future[ValveSwitch]` as a materialized value of `Valve`. That future should be completed when it is safe to invoke async callbacks which is what ValveSwitch exposes.

cchantep:
@2m I doubt I would have time to wrap my head around such internal change, so if you could at least draft something about?

patriknw:
what is the motivation for updating to 2.5.0? If we continue building with 2.4.x it can be used with both 2.4.x and 2.5.x, but otherwise it would only be usable with 2.5.x

cchantep:
Because the current 2.4 is not usable to test Akka 2.5 based project

patriknw:
then something in akka-stream-contrib must be using internal APIs of akka-stream, might be worth investigating that instead

cchantep:
The Akka `MaterializerImpl` as in the current changes : https://github.com/akka/akka-stream-contrib/pull/94/files#diff-e3e91fe15856c843a3a6db5e51467e9aL24

2m:
Hi @cchantep. This PR uncovered a bigger issue where akka-stream-contrib is not tested against Akka 2.5 and by now there have been a couple of issues accumulated. I have created a https://github.com/akka/akka-stream-contrib/pull/96 to fix that which supersedes your PR. Thank you for raising this issue.

PR #96, opened 2017-05-09
Run tests with Akka 2.5

2m:
Needed a couple of fixes for the tests to pass with Akka 2.5:

* removed dependence on the `ActorMaterializerImpl` from `assertAllStagesStopped`. As it only needs a `supervisor` - `ActorMaterializer` is enough.
* new materializer in Akka 2.5 uncovered timing issues in Valve graph stage, where async callback is invoked before Valve stage finishes initializing. Changed stages materialized value to a `Future[...]` instead. It would be best to later fix the underlying issue https://github.com/akka/akka/issues/20503#issuecomment-218455275

Supersedes #94


project/Common.scala
```diff
@@ -11,7 +11,10 @@ import com.typesafe.sbt.SbtScalariform.ScalariformKeys
 
 object Common extends AutoPlugin {
 
-  val AkkaVersion = "2.4.12"
+  val AkkaVersion = sys.env.get("akka") match {
+    case Some("2.5") => "2.5.1"
+    case _ => "2.4.18"
+  }
```
ktoso: wasn't too bad :)

Can we do `akka.version` instead?

ktoso:
*Who is that guy?!* ;-)

2m:
There used to be `akka.version` in Akka build, but that one would set the version of Akka itself. But I might not have build definitions swapped in of other projects as well.

I went with `akka` as it was nice and clean. And it also does not store full version but just series. I did not want to set full version in the env variable as then it needs to be duplicated: once in the travis.yml and then second time in build.sbt for a default value.

johanandren:
Ok, apology accepted! ;)

ktoso:
Still, calling it `akka.version` is more understandable when someone looks at this IMO,
and we may also want to at some point use `akka.sources` as https://github.com/akka/akka-http/blob/master/project/AkkaDependency.scala#L11 does so it would be clearer which one is which

2m:
`akka.series`?

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -50,27 +50,30 @@ object Valve {
...
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[ValveSwitch]) = {
     val logic = new ValveGraphStageLogic(shape, mode)
-    (logic, logic.switch)
+    (logic, logic.promise.future)
   }
 
   private class ValveGraphStageLogic(shape: Shape, var mode: SwitchMode) extends GraphStageLogic(shape) with InHandler with OutHandler {
-    val switch = new ValveSwitch {
+
+    val promise = Promise[ValveSwitch]
```
patriknw: ok, I guess this is the same as issue https://github.com/akka/akka/issues/20503
`Future` is not so nice, but we can't do much else until that ticket is fixed

project/Common.scala
patriknw: ok, so this is testing source compatibility, but not binary compatibility. better than nothing

ktoso:
`akka.series` +1 :-)

Issue #116, opened 2017-11-07
Facility to obtain the state of a valve's switch


zaharidichev:
Currently the `ValveSwitch` looks like 

```scala
sealed trait ValveSwitch {
  /**
   * Change the state of the valve
   *
   * @param mode expected mode to switch on
   * @return A future that completes with true if the mode did change and false if it already was in the requested mode
   */
  def flip(mode: SwitchMode): Future[Boolean]
}
```

I propose to add another method that will return a Future[SwitchMode] telling us what the state of the valve at the moment is. We have a strong case for using that in production.

ktoso:
Feel free to submit a PR, sounds useful

PR #118, opened 2017-11-09
Add .getMode to the the valve swich to be able to check the current state

zaharidichev:
As described in #116 there is a getMode method on the `ValveSwitch` now.

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/Valve.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/Valve.scala
@@ -20,6 +20,13 @@ sealed trait ValveSwitch {
    * @return A future that completes with true if the mode did change and false if it already was in the requested mode
    */
   def flip(mode: SwitchMode): Future[Boolean]
+
+  /**
+    * Obtain the state of the valve
+    *
+    * @return A future that completes with [[SwitchMode]] to indicate the current state of the valve
+    */
+  def getMode(): Future[SwitchMode]
 }
 
 object Valve {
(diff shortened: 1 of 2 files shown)


contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -98,9 +105,17 @@ final class Valve[A](mode: SwitchMode) extends GraphStageWithMaterializedValue[F
...
       override def flip(flipToMode: SwitchMode): Future[Boolean] = {
         val promise = Promise[Boolean]()
-        callback.invoke((flipToMode, promise))
+        flipCallback.invoke((flipToMode, promise))
+        promise.future
+      }
+
+      override def getMode(): Future[SwitchMode] = {
+        val promise = Promise[SwitchMode]()
+        getModeCallback.invoke(promise)
```
johanandren: Note that this will never complete if the stage was already stopped, or stops before the async callback gets invoked. There is a new `invokeWithFeedback()` that recently got merged in Akka, but is not yet released that solves the problem.

zaharidichev: Thanks for the tip @johanandren, very useful. So what do you suggest ? We can wait on this one until `invokeWithFeedback` is released

johanandren: Maybe put a `// FIXME will never complete promise if stage is stopped, use invokeWithFeedback when Akka 2.5.7 is released` in here at least

zaharidichev: done

contrib/src/main/scala/akka/stream/contrib/Valve.scala
```diff
@@ -98,9 +105,18 @@ final class Valve[A](mode: SwitchMode) extends GraphStageWithMaterializedValue[F
           promise.success(succeed)
       }
 
+      // FIXME will never complete promise if stage is stopped, use invokeWithFeedback when Akka 2.5.7 is released
```
raboof: Maybe it'd make sense to also already create a github issue to keep track of this... but this is probably sufficient for now :)

zaharidichev: will make sure to follow up on next release

PR #120, opened 2017-12-13
Allow passing timeout parameter to unfoldFlow

liff:
Instead of unconditionally loading it from global configuration.


contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -7,27 +7,35 @@ import akka.stream._
...
 import akka.stream.stage._
 import com.typesafe.config.ConfigFactory
-import scala.concurrent.duration.Duration
+import scala.concurrent.duration.{ Duration, FiniteDuration }
 
 /**
  * Source factory methods are placed here
  */
 object SourceGen {
+  private[this] lazy val defaultTimeout = Duration.fromNanos(ConfigFactory.load().getDuration("akka.stream.contrib.unfold-flow-timeout").toNanos)
```
johanandren: Loading the config separately is not a good idea, it should come from the actor system. I see two reasonable options:
1. Have the stage/logic take an option, and if that is empty get the system.settings.config from the running stream
2. Just skip the config part entirely (because there is no reasonable way to have a default as it depends on the application)

raboof: While I realize it was already done like this in `UnfoldFlow.scala`, we shouldn't use `ConfigFactory.load()` like this. Also it would be better to use the [`getDuration` with 2 parameters](https://lightbend.github.io/config/latest/api/com/typesafe/config/Config.html#getDuration-java.lang.String-java.util.concurrent.TimeUnit-).

contrib/src/main/scala/akka/stream/contrib/SourceGen.scala
```diff
@@ -7,27 +7,35 @@ import akka.stream._
...
+   * If inner flow fails to complete/fail downstream, stage is
+   * failed with an IllegalStateException.
+   *
    * IMPORTANT CAVEAT:
    * The given flow must not change the number of elements passing through it (i.e. it should output
    * exactly one element for every received element). Ignoring this, will have an unpredicted result,
    * and may result in a deadlock.
    */
-  def unfoldFlow[S, E, M](seed: S)(flow: Graph[FlowShape[S, (S, E)], M]): Source[E, M] = {
+  def unfoldFlow[S, E, M](seed: S, timeout: FiniteDuration)(flow: Graph[FlowShape[S, (S, E)], M]): Source[E, M] = {
```
raboof: To make this possible and convenient at least for the Scala API we could make this timeout parameter implicit. Also it would be good to use the `Timeout` type here.

liff:
I agree that doing `ConfigFactory.load()` in library code is somewhat imploite, I kept it in interest of not (potentially) breaking existing code. If that is not a concern, then I'd certainly be in favor of removing the timeout-parameterless method altogether.

So, do we get it from `system.settings` or as an implicit `Timeout`?

johanandren:
I'd vote for implicit timeout. I don't think it is a big problem that we would not keep existing behaviour here as it is easy to workaround by just reading that config key manually and this would be source incompatible anyway so users will notice the change.

liff:
@raboof is this ok for you now?

raboof:
@liff sorry to keep this around for a while, it looks like this PR has since developed a merge conflict. Could you fix it up so we can merge?

liff:
Conflicts fixed.

I had removed `reference.conf` but looks like #87 wants to use `ConfigFactory.load()` again...

liff:
Looks like the CI build failed on an unrelated file. I included its reformatting in this PR to get things moving, hopefully.

Issue #123, opened 2018-02-08
Release 0.9


milandesai:
I wanted to use the DelayFlow that was added last July, but the latest release is from last May. Given that it's been 9 months, can we do a 0.9 release?

ennru:
0.9 has been released a while ago now: https://github.com/akka/akka-stream-contrib/releases/tag/v0.9

PR #124, opened 2018-02-08
Deploy from travis

raboof:
Hmm, looks like we need to add a way to set the sonatype credentials via
environment variables? And document that you'll have to close the repo
from the sonatype UI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,11 @@ cache:
   directories:
     - $HOME/.ivy2/cache
     - $HOME/.sbt/boot
+
+deploy:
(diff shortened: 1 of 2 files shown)


raboof:
refs #123

2m:
Yup. We need to encrypt `SONA_USER` and `SONA_PASS` and then we can call `bintraySyncMavenCentral` from after publish from travis as well. So there will be no need to go to bintray ui, neither to sonatype ui.

I have a configured project here: https://github.com/2m/authors/blob/master/.travis.yml

PR #126, opened 2018-02-14
Consolidate the format of README

2m:
So it has the same badge markup format as Alpakka and others.


2m:
It has been a long time since I've seen something so breathtaking as well!

PR #142, opened 2018-10-29
Add PassThrough Flow

tkroman:
As discussed here: https://discuss.lightbend.com/t/passing-stream-elements-into-and-over-flow/2536

cc @2m

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/PassThroughFlow.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/PassThroughFlow.scala
@@ -95,7 +95,7 @@ object PassThroughFlow {
         builder.add(ZipWith[I, O, R]((left, right) => output(left, right)))
 
       // format: off
-      broadcast.out(0)          ~>          zip.in0
+      broadcast.out(0) ~> zip.in0
       broadcast.out(1) ~> processingFlow ~> zip.in1
       // format: on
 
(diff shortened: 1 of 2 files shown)


contrib/src/main/scala/akka/stream/contrib/PassThroughFlow.scala
```diff
@@ -0,0 +1,105 @@
...
+package akka.stream.contrib
+
+import akka.NotUsed
+import akka.stream._
+import akka.stream.scaladsl._
+
+/**
+ * Given `Flow[In, Out]` and `f: (In, Out) => Result`,
+ * transforms original flow into a `Flow[In, Result]`, i.e.
+ * for every element `e` signalled from upstream, emits `f(e, flow(e))`.
```
johanandren: Very important that the flow is 1:1, and does no re-ordering, as soon as it drops, creates more elements or re-orders the downstream be incorrect. (Which is the reason there is no built in tool for it, we have discussed it over and over again to try to figure out solutions though, see https://github.com/akka/akka/issues/15957 and linked issues)

tkroman: Right, that's what I also gathered from the links in the original discussion. Do you want me to make this more explicit in docs or do you mean that since it's not a generic solution, it shouldn't be here?

johanandren: I think it needs to be a *blink UPPERCASE* warning in the top of the docs but ok in contrib if that is added.

tkroman: done

PR #143, opened 2018-10-29
Add by-name version of Source.repeat

tkroman:
People often need this and I think there is an advantage to having this as opposed to always doing

```scala
Source.repeat(()).map(_ => ...)
```


johanandren:
Oh wait, I approved too quickly, some test coverage would be good.

tkroman:
there are some basic tests, if you have any suggestions on what can be expanded, i'd be happy to do that.

johanandren:
I must have missed them somehow. Looks good!

Issue #144, opened 2018-11-15
Add maxNumber and maxTime to AccumulateWhileUnchanged


LysanderGG:
`AccumulateWhileUnchanged` only emits when the property changes, which can be an issue if the property doesn't change very often. It would be similar to `groupedWithin`, becoming `Chunk up this stream into groups of elements received within a time window, or limited by the number of the elements, or by the property changing, whatever happens first.`

I already customized `AccumulateWhileUnchanged` for a project, so I can open a PR if someone else thinks it can be useful.

PR #145, opened 2018-11-16
Add maxElements and maxDuration parameters to AccumulateWhileUnchanged

LysanderGG:
See issue #144 

Any feedback is welcome

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -82,12 +82,9 @@ final class AccumulateWhileUnchanged[Element, Property](
         if (currentState.isEmpty) currentState = Some(nextState)
 
         (currentState, maxElements) match {
-          case (Some(`nextState`), max) if max.isEmpty || nbElements < max.get =>
-            buffer += nextElement
-            nbElements += 1
-            pull(in)
-          case (_, _) =>
-            pushResults(Some(nextElement), Some(nextState))
+          case (Some(`nextState`), None) => stash(nextElement)
+          case (Some(`nextState`), Some(max)) if nbElements < max => stash(nextElement)
+          case _ => pushResults(Some(nextElement), Some(nextState))
         }
       }
 
@@ -103,6 +100,12 @@ final class AccumulateWhileUnchanged[Element, Property](
           emit(out, result)
         }
         completeStage()
+      }
+
+      private def stash(nextElement: Element) = {
+        buffer += nextElement
+        nbElements += 1
+        pull(in)
       }
     })
 
@@ -136,7 +139,7 @@ final class AccumulateWhileUnchanged[Element, Property](
 
       nextElement match {
         case Some(next) =>
-          buffer += nextElement.get
+          buffer += next
           nbElements += 1
         case None => Unit
       }
--- a/contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -5,10 +5,11 @@
 package akka.stream.contrib
 
 import akka.japi.function
-import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.stage.{ GraphStage, TimerGraphStageLogic, InHandler, OutHandler }
 import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
 
 import scala.collection.immutable
+import scala.concurrent.duration.FiniteDuration
 
 object AccumulateWhileUnchanged {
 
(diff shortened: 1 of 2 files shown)


contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -63,19 +83,22 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
 
         currentState match {
           case Some(`nextState`) =>
-            buffer += nextElement
-            pull(in)
+            if (maxElements.isDefined && nbElements >= maxElements.get) {
```
2m: `maxElements` is an option, so this `if/else` can be rewritten to:

```scala
maxElements match {
  case Some(max) if nbElements >= max => pushResults(...)
  case _ =>
    ...
}
```
Or, even better, it can be flattened to the enclosing match of the `currentState`.

contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -87,8 +110,38 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
       }
     })
 
+    override def preStart(): Unit = {
+      super.preStart()
+      if (maxDuration.isDefined) {
+        schedulePeriodically(None, maxDuration.get)
+      }
```
2m: Convert to `match` expression as well.

contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -87,8 +110,38 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
...
+      nbElements = 0
+
+      if (result.nonEmpty) {
+        push(out, result)
+      }
+
+      if (nextElement.isDefined) {
+        buffer += nextElement.get
+        nbElements += 1
+      }
```
2m: Use `match` here as well.

contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
LysanderGG: The best I can think of with one match is:

```
(currentState, maxElements) match {
  case (Some(`nextState`), max) if max.isEmpty || nbElements < max.get =>
    buffer += nextElement
    nbElements += 1
    pull(in)
  case (_, _) =>
    pushResults(Some(nextElement), Some(nextState))
}
```

What do you think?

lightbend-cla-validator:
Hi @LysanderGG,

lightbend-cla-validator:
Hi @LysanderGG,

LysanderGG:
@2m Thank you for the review!
 I pushed some changes, please have a look. I'm not sure about the match to replace a single `if`.

Note: I signed the CLA.

contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -137,10 +134,13 @@ final class AccumulateWhileUnchanged[Element, Property](
         push(out, result)
       }
 
-      if (nextElement.isDefined) {
-        buffer += nextElement.get
-        nbElements += 1
+      nextElement match {
+        case Some(next) =>
+          buffer += nextElement.get
```
2m: You can use `next` here instead of `nextElement.get`

contrib/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
```diff
@@ -61,21 +81,20 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
 
         if (currentState.isEmpty) currentState = Some(nextState)
 
-        currentState match {
-          case Some(`nextState`) =>
+        (currentState, maxElements) match {
+          case (Some(`nextState`), max) if max.isEmpty || nbElements < max.get =>
```
2m: I would like to avoid calling `.get`. It is an unsafe operation (can throw an exception). It is only safe to call it depending on the context. Here it is currently fine, since we check that `max` is not empty, but such code does not age well during refactorings.

How about extracting the following to a private method called for example `stash`:
```scala
buffer += nextElement
nbElements += 1
pull(in)
```

and then calling that from:
```scala
(currentState, maxElements) match {
  case (Some(`nextState`), None) => stash(nextElement)
  case (Some(`nextState`), Some(max)) if nbElements < max => stash(nextElement)
  case _ => pushResults(Some(nextElement), Some(nextState))
}
```

LysanderGG: Thank you for your detailed explanation!
I fixed it as suggested.

PR #150, opened 2019-02-28
PartitionWith: add eagerCancel parameter.

TomasMikula:
When one downstream of `PartitionWith` cancels, events to the other downstream should keep flowing. This is consistent with what `Partition` does.

### DIFF
--- a/contrib/src/main/scala/akka/stream/contrib/PartitionWith.scala
+++ b/contrib/src/main/scala/akka/stream/contrib/PartitionWith.scala
@@ -18,23 +18,29 @@ object PartitionWith {
    * Factory for [[PartitionWith]] instances.
    *
    * @param p partition function
+   * @param eagerCancel when `false` (the default), cancel after all downstream have cancelled.
+   *                    When `true`, cancel as soon as any downstream cancels and complete the remaining downstreams
    * @tparam In input type
    * @tparam Out0 left output type
    * @tparam Out1 right output type
    * @return [[PartitionWith]] instance
    */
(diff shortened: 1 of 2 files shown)


johanandren:
In fan out stages we usually call this "eager cancellation" and allow it to be chosen by the user upon creation.

TomasMikula:
I see. I will add `eagerCancel: Boolean` constructor parameter and then update this PR.

TomasMikula:
Updated.

contrib/src/main/scala/akka/stream/contrib/PartitionWith.scala
```diff
@@ -120,6 +133,18 @@ final class PartitionWith[In, Out0, Out1] private (p: In => Either[Out0, Out1])
...
+      override def onDownstreamFinish(): Unit =
+        downstreamFinished()
     })
+
+    private def downstreamFinished(): Unit = {
+      activeDownstreamCount -= 1
+      if (eagerCancel) {
+        completeStage()
+      } else if (activeDownstreamCount == 0) {
+        cancel(in)
```
johanandren: Shouldn't it be `completeStage()` here as well?

TomasMikula: Does it make any difference? By this time, both downstreams have finished, so it remains to only signal cancellation to the upstream.

johanandren: Nah, maybe not.

johanandren:
Fixes #150

johanandren:
Lol, 150 is this PR not an attached issue. NVM.

PR #151, opened 2019-03-01
Update versions.

TomasMikula:


TomasMikula:
Any suggestion to fix the error

```
[error] /home/travis/build/akka/akka-stream-contrib/contrib/src/main/scala/akka/stream/contrib/TestKit.scala:42:53: value PrintDebugDump is not a member of object akka.stream.impl.StreamSupervisor
[error] Error occurred in an application involving default arguments.
[error]               children.foreach(_ ! StreamSupervisor.PrintDebugDump)
[error]                                                     ^
```

in a way that works with Akka 2.4.20 as well as 2.5.21?

Can use of internal Akka API be avoided?

Is it reasonable to expect that the same source code will work with Akka 2.4.x and Akka 2.5.x?

raboof:
> Is it reasonable to expect that the same source code will work with Akka 2.4.x and Akka 2.5.x?

Roughly speaking, we [keep binary compatibility of public API's](https://doc.akka.io/docs/akka/2.5/common/binary-compatibility-rules.html) - so a library compiled against Akka 2.4.x should work in a project that puts Akka 2.5.x on the classpath instead of 2.4.x. This usually also means source compatibility: usually, the same source code that worked for Akka 2.4.x should also work for 2.5.x - but not always.

This does not go for `StreamSupervisor.PrintDebugDump`, however, since that API is marked 'Internal', and we only keep compatibility for public API's.

Since this is a contrib library, it might be reasonable to just use the new `MaterializerState` and not keep binary compatibility here - we should then make that clear in the documentation, though. WDYT?

TomasMikula:
@raboof But currently `akka-stream-contrib` is built for both Akka 2.5.x and 2.4.x, and `MaterializerState` is not available in Akka 2.4.x. Should that be changed to only build with the latest Akka.

raboof:
Ah I did not realize that. I think it would be OK to drop support for Akka 2.4 in this library, given 2.5 has now been out for almost 2 years, but that probably warrants a separate discussion/PR.

2m:
I am cleaning up akka-stream-contrib build for preparations of snapshot publishing. Will incorporate these proposed changes there.

TomasMikula:
👍

PR #153, opened 2019-03-04
Drop support for Akka 2.4.x.

TomasMikula:
No longer require sources to compile with Akka 2.4 series.

@raboof in #151:

> I think it would be OK to drop support for Akka 2.4 in this library, given 2.5 has now been out for almost 2 years

Generally, it seems OK to me to require some minimal Akka version for each akka-stream-contrib version, as thanks to Akka's backward compatibility users should be able to safely update their Akka version as well.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -15,7 +15,7 @@ object Common extends AutoPlugin {
 
   val AkkaVersion = sys.env.get("AKKA_SERIES") match {
     case Some("2.5") => "2.5.13"
-    case _ => "2.4.18"
+    case _ => "2.5.13"
   }
 
   override def trigger = allRequirements
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ deploy:
 
 env:
   matrix:
-    - AKKA_SERIES=2.4
     - AKKA_SERIES=2.5
   global:
     - BINTRAY_USER=raboofje


project/Common.scala
```diff
@@ -15,7 +15,7 @@ object Common extends AutoPlugin {
 
   val AkkaVersion = sys.env.get("AKKA_SERIES") match {
     case Some("2.5") => "2.5.13"
-    case _ => "2.4.18"
+    case _ => "2.5.13"
   }
```
TomasMikula: This whole `match` expression could be replaced by `"2.5.13"`, but I wanted to preserve the logic in case we support multiple `AKKA_SERIES` in the future (say 2.5 and 2.6).

TomasMikula: Although if my reasoning that it is OK to require some specific minimal Akka version for each akka-stream-contrib release is correct, then we can get rid of this `AKKA_SERIES` environment variable for good.

2m: Because of the binary compatibility guarantees, it is possible to depend on the latest version from the series and use the latest features and bugfixes. Binary compatibilty guarantees that older code will work fine with a later Akka version without recompiling.

I say lets get rid of the env variable all together for now, and use the latest Akka version.

patriknw: That statement is wrong or at least misleading. Binary backwards compatibility means that an application using this library can use the latest Akka version 2.5.21 even though this library was built with Akka 2.5.13. The opposite is not true. If this library was built with Akka 2.5.21 an application is not guaranteed to be able to run with an older version of Akka, like 2.5.20. In other words, the version defined here specifies the minimum required version.

For akka-stream-contrib I don't have any opinions about what version that is supported. It's perfectly fine to use latest Akka (and definitely no need to support 2.4.x).

TomasMikula: Right. So any time akka-stream-contrib bumps up its minimal required Akka version, users will have to bump up their Akka version as well. That, however, should not cause them any trouble.

2m: Yea, sorry for being a bit vague here. Thanks @patriknw for clearing it up.

PR #155, opened 2019-03-08
Add Jdk-9 using Automatic module

Raghav2211:


lightbend-cla-validator:
Hi @rusty-neo,

Raghav2211:
> Thanks for noticing this is missing!
> 
> The modules from akka itself have names like 'akka.actor', 'akka.cluster.metrics' and 'akka.stream.testkit'. Perhaps we should make this 'akka.stream.contrib'? Even though this library is not in the akka repository, it is still in the akka organization, so that might be more consistent?

Yes this is consistent for now because **akka-stream-contrib**  is packaging separately and in akka repo there is class called **AutomaticModuleName** which is converting all aka modules to automatic modules, when this library merge in akka then it will be build as a akka module and will be take care of their's  build.sbt.
So it think this is consistent for now in next release

raboof:
We also put alpakka-kafka under the `akka.stream.` prefix though (https://github.com/akka/alpakka-kafka/blob/master/build.sbt#L134), and for that (like contrib) we also don't have any plans to move them into the Akka repo. What are the implications of `Automatic-Module-Name`? Are there any implications to the structure of the name chosen?

Raghav2211:
> We also put alpakka-kafka under the `akka.stream.` prefix though (https://github.com/akka/alpakka-kafka/blob/master/build.sbt#L134), and for that (like contrib) we also don't have any plans to move them into the Akka repo. What are the implications of `Automatic-Module-Name`? Are there any implications to the structure of the name chosen?

No implication if you add any library, this change will only able to add this module in module-info.java where we need.

project/Common.scala
```diff
@@ -24,7 +25,8 @@ object Common extends AutoPlugin {
   override lazy val projectSettings = SbtScalariform.scalariformSettings ++ Seq(
     organization := "com.typesafe.akka",
     organizationName := "Lightbend Inc.",
-
+    packageOptions := Seq(ManifestAttributes(
+                      ("Automatic-Module-Name", "akka-stream-contrib"))),
```
ennru: I think we'd like to go with `akka.stream.alpakka.contrib` instead.

raboof:
(closing due to inactivity, feel free to reopen when work resumes!)

PR #164, opened 2019-06-19
Document travis now also does the close+release+drop

raboof:


raboof:
Thanks @2m, great work here!

PR #169, opened 2019-07-20
Fix token throttle completion issue and improve token pre-fetching

### LINKED ISSUE #168: TokenThrottle can fail to complete and doesn't consistently pre-fetch tokens
While reviewing the TokenThrottle code I submitted in #167 , I noticed that it has two issues.

First, the stage would not complete if:
* an element was buffered 
* tokens have been prefetched (i.e. `tokens > 0`)
* too few tokens were available to emit the element (i.e. `tokens < cost`).

Also, the stage would not consistently pre-fetch tokens if the downstream did not yet pull, but the currently buffered element was too expensive to emit with the currently available token budget. It also might defer completions triggered by the token source.

PR will follow suit.

andreas-schroeder:
Fixes #168

Stage now completes if an element was buffered and tokens were available, but too few to emit the element, and the token source would complete.

Stage now pre-fetches tokens if the downstream did not yet pull, but the currently buffered element is too expensive to emit with the currently available token budget.

### DIFF
--- a/src/main/scala/akka/stream/contrib/TokenThrottle.scala
+++ b/src/main/scala/akka/stream/contrib/TokenThrottle.scala
@@ -77,25 +77,34 @@ final class TokenThrottle[A](costCalculation: A => Long) extends GraphStage[FanI
     var elemsCompleted = false
 
     private def maybeEmit(): Unit =
-      if (isAvailable(out) && buffer.isDefined) {
+      if (buffer.isDefined) {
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

PR #170, opened 2019-07-25
add ZipToStreamFlow

kirked:
Adds a facility for predictable-memory streaming zipfile creation, as in the instance of sending multiple files to a browser over an HTTP connection (even when they originate from a remote service such as S3).

Incoming InputStreams must be able to be created synchronously, but are delayed such that only a single InputStream is open at any given time.

Original gist at https://gist.github.com/kirked/03c7f111de0e9a1f74377bf95d3f0f60 (original license MIT).

The author is happy to contribute this and derivative works for inclusion in the akka-stream-contrib library, and hereby assigns full control over this code, including copyright protection and licensing worldwide, in perpetuity, to Lightbend, Inc.

## References

References #xxxx

### DIFF
--- a/src/main/scala/akka/stream/contrib/ZipToStreamFlow.scala
+++ b/src/main/scala/akka/stream/contrib/ZipToStreamFlow.scala
@@ -1,0 +1,188 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.contrib
+
+import akka.stream._
+import akka.stream.scaladsl._
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler, StageLogging}
+import akka.util.{ByteString, ByteStringBuilder}
+import java.io.{InputStream, OutputStream}
+import scala.util.control.NonFatal
+
+/**
+ * Streamed zip-file creation Akka streams implementation.
+ * Typical usage, assuming an implicit ActorSystem:
+ *
+ *     import akka.stream._
+ *     import akka.stream.scaladsl._
+ *
+ *     val filesToZip: Iterable[ZipToStreamFlow.ZipSource] = ...
+ *     Source(filesToZip)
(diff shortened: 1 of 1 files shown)


src/main/scala/akka/stream/contrib/ZipToStreamFlow.scala
```diff
@@ -0,0 +1,187 @@
...
+            }
+          }
+
+        result.iterator.filter(_.nonEmpty)
+      }
+    }
+}
+
+object ZipToStreamFlow {
+  final case class ZipSource(filePath: String, streamGenerator: () => InputStream)
```
raboof: It might be nicer to use Akka streams in this API (even if they'd eventually need to be converted to `InputStream` - WDYT?

kirked: Yes. When I wrote this, our need was for zipping multiple S3-contained assets to the browser, and the AWS API (at least that we were using) provided synchronous access to the InputStreams. There was no Alpakka, and Akka was just joining our tech stack (I still have the original Play Iteratee version as a gist).

I see this as a good start for an often-needed utility that can be greatly improved.

2m: I also prefer the implementation to use Akka Streams API. Convertion from Java stream to Akka Stream is always possible with `StreamConverters`: https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/StreamConverters.scala#L25

2m:
This is useful, but as discussed, it would be better to implement this functionality using Akka Stream APIs.

I am going to close this PR here, but the way forward for Zip streams would be to move the implementation of #171 to Alpakka File connector and then enrich that with some examples of Java Streams and using `akka.stream.scaladsl.StreamConverters`. That should cover the same use-cases as are solved in this PR.

@kirked feel free to move the #171 to Alpakka and work together with @michalbogacz on this.

PR #171, opened 2019-07-28
Flow to merge streams to ZIP file

michalbogacz:
## Purpose

Flow to merge streams to ZIP file.
My usage was to download files from S3 and pass them to the user as ZIP file.

Similar to:
https://github.com/akka/akka-stream-contrib/pull/170
But does not requires InputStream as input (and blocking a thread for creating them https://doc.akka.io/docs/akka/current/stream/operators/index.html#additional-sink-and-source-converters).

Base on my work in:
https://gist.github.com/michalbogacz/08868f51b7053cadfef00f51196bf2ec

### DIFF
--- a/src/main/scala/akka/stream/contrib/MergeToZipFile.scala
+++ b/src/main/scala/akka/stream/contrib/MergeToZipFile.scala
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.contrib
+
+import java.util.zip.{ZipEntry, ZipOutputStream}
+import akka.NotUsed
+import akka.stream._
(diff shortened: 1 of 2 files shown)


ennru:
This would be relevant to be added to Alpakka instead. See https://github.com/akka/alpakka/issues/658

michalbogacz:
@ennru I didn't know there is an issue for this. I think I can prepare PR to Alpakka but it will take me some time

2m:
I think this can go to the `file` (https://github.com/akka/alpakka/tree/master/file) Alpakka connector. That will make it easier to create Alpakka PR than to create a new connector all-together.

2m:
I am going to close this now, as we prefer to have this in Alpakka. We are really looking forward to having a PR in the Alpakka repo.

PR #175, opened 2019-10-11
MergeByIndex fan-in stage

### LINKED ISSUE #174: Support for sorting fan-in shape with more than two upstreams
When distributing elements over multiple partitions, it is sometimes useful to merge them back in an order-preserving way - for example to realize a conservative checkpointing and by this proving at-least-once processing guarantees. 
While it is possible to use `zipWithIndex` to create a total ordering over the items and partition them to multiple branches, the `MergeSorted` fan-in stage can only merge two upstreams. For merging multiple upstreams, a tree of `MergeSorted` stages needs to be constructed.

What is missing in Akka Stream is a more convenient fan-in stage that
- is able to preserve total ordering of elements given by e.g. `zipWithIndex` stages,
- can merge an arbitrary number of upstreams, and
- can cope with index omissions created by using e.g. `filter` stages.

andreas-schroeder:
This pull requests proposes a new `MergeByIndex` stage for merging multiple upstreams by index (i.e. by total order). It supports a custom total order over incoming elements, and offers a convenience constructor for tuples created via `.zipWithIndex`. The stage requires a total ordering, and as such, fails fast as soon as it detects that the index is not strict monotonically increasing. `MergeByIndex` can handle index omissions that can be created by upstreams filtering out elements that shouldn't be further processed. It doesn't yet support eager completion; currently, it will complete only after all its upstreams completed and internally buffered elements are emitted.

Internally, `MergeByIndex` works by keeping track of the next expected index, starting at zero. It initially pulls from all its upstreams, and buffers fetched elements using a priority queue if they don't have the expected index. On index omission, the next expected index will however never arrive. This case can be handled by the stage once full information is available, i.e. once all upstreams have published an element. Once this happened without the expected index appearing,
it is safe to emit the element with the lowest index available. The logic for handling index omissions is a bit more elaborate to account for potential upstream completions, but the underlying principle remains the same.

Using a priority queue for buffering causes O(log inputPorts) added time complexity per element; note that this is also the best one can achieve with a well-balanced tree of `MergeSorted` stages.

Fixes #174

### DIFF
--- a/src/main/scala/akka/stream/contrib/MergeByIndex.scala
+++ b/src/main/scala/akka/stream/contrib/MergeByIndex.scala
@@ -98,27 +98,27 @@ final class MergeByIndex[T](val inputPorts: Int, index: T => Long) extends Graph
       if (buffer.nonEmpty) {
         if (buffer.head._2 == expectedIndex) {
           emitAndPull(buffer.dequeue())
-        } else if (dataFromAllInletsBuffered) {
-          // if all inlets pushed data and we didn't find the expected index, we know this is an index omission.
+        } else if (elementsFromAllInletsBuffered) {
+          // if all inlets pushed elements and we didn't find the expected index, we know this is an index omission.
           // it is therefore fine and necessary to emit the element with the smallest index seen.
           emitAndPull(buffer.dequeue())
         }
       }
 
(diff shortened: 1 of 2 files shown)


ennru:
Thank you!

PR #176, opened 2019-10-26
Add TarballFlow

choffmeister:
In one of my personal projects I found the need to generate a simple tarball from a `Source[(String, Long, Source[ByteString, _])]`. I am not sure if there is enough demand for this to include it into this library here. This PR is at the moment just a simple copy paste - so no sophisticated testing.

If there is enough demand, I would be willing to improve the tests and do whatever else is needed to have it merged. So this is basically an offer with some preview code that - if wanted - would need some more work from my side.

I would love to hear feedback.

### DIFF
--- a/src/main/scala/akka/stream/contrib/TarballFlow.scala
+++ b/src/main/scala/akka/stream/contrib/TarballFlow.scala
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.contrib
+
+import java.lang.Long.toOctalString
+import java.time.Instant
+
+import akka.NotUsed
+import akka.stream.scaladsl.{Flow, Source}
+import akka.util.ByteString
+
+final case class TarballHeader(fullFileName: String, fileSize: Long, lastModification: Instant = Instant.now) {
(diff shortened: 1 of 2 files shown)


ennru:
Hi @choffmeister,
Thank you for this suggestion to add this here! I think this would be nice to have in the Alpakka file connector which only recently got [Zip support](https://doc.akka.io/docs/alpakka/snapshot/file.html#zip-archive) (yet unreleased).

choffmeister:
@ennru Nice to hear. Just out of curiosity: In my project I tried ZIP archives first with the approach to uses Java's `ZipOutputStream` in combination with akka's `StreamConverters.asOutputStream`. Even though I disabled compression I had massive CPU load when generating the ZIP archive. With my own Tarball implementation that does not use old Java IO streams the CPU load is minimal.

Anyway: Please let me know where I my PR needs to be improved to get to a mergeable state.

seglo:
I second @ennru 's suggestion of contributing this to the Alpakka File connector (instead of akka-stream-contrib).  I believe it would get more visibility and usage there.

choffmeister:
Nice, didn't know that there is a File connector in alpakka. Yeah this might make more sense. Will close this and open a PR there.

PR #177, opened 2019-10-27
BufferedBroadcast - if one output consumes elements, it keeps buffering them for the other outputs

amorfis:
## Purpose

Introduces new graph stage - BufferedBroadcast

### DIFF
--- a/src/main/scala/akka/stream/contrib/BufferedBroadcast.scala
+++ b/src/main/scala/akka/stream/contrib/BufferedBroadcast.scala
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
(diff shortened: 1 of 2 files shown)


ennru:
Thank you for this suggestion, sorry for not getting to it earlier.

Does this behave differently to a regular broadcast with a buffer operator on downstreams?

amorfis:
I'm not sure how regular broadcast with buffers work. Does the buffer automatically pull elements? If so, it seems like it behaves the same. 

I created it to prevent a downstream from "starving", when it needed to pull elements from broadcast, but other downstream was not pulling. Broadcast then did not push, because it does so only when all the downstreams pull.

ennru:
Yes, that seems to be solvable with regular `buffer` operators.

We always recommend the composition of existing operators over creating special implementations for this kind of use cases.

PR #179, opened 2019-11-19
Update sonatype credentials

raboof:
to publish snapshots and releases


raboof:
fixed failure in #180

PR #182, opened 2020-01-29
Update release status (no longer released)

ignasi35:


zackman0010:
Does this mean that there will no longer be any commits to this repo? If there are future commits here, are we intended to download the source and compile it ourselves to use anything from this repo?

ignasi35:
> Does this mean that there will no longer be any commits to this repo? If there are future commits here, are we intended to download the source and compile it ourselves to use anything from this repo?

There can be more commits, the repo is not archived. As you guessed, users are free to take this repo and publish artifacts on their infrastructure.

zackman0010:
Thank you!
Just out of curiosity and if you don't mind me asking, why was the decision made to go this way instead of continuing to make releases for this?

raboof:
Summarizing, being an artifact published under `com.typesafe.akka` made people expect the code in this module to be 'akka quality' (despite the '-contrib' name), which in turn made us reluctant to merge contributions without in-depth code review.

Because part of the reason to have akka-stream-contrib at all is be to have a place where we can be more quick to merge contributions, this kind of defeated the purpose. We hope by making it a 'somewhat tested collection of contributed components' rather than a 'production artifact published by typesafe/Lightbend' this would get more low-friction again.

PR #183, opened 2020-02-20
Introduce LatencyTimer operator.

lostiniceland:
LatencyTimer provides a operator which can wrap any Flow-Shape and measure how long each element took.
The operator uses the following shape
I -> O, TimedResult[O] when wrapping a Flow[I,O]

The TimedResult contains the wrapped flows result as well in case further logic is needed. Example would be to distinguish between different metrics based on the result of the wrapped flow.

The operator does only support flows which are preserving the order, so no async should be done inside (only mapAsync).

I could not come up with a better name than LatencyTimer and measureLatency in the DSL. Basically "Timed" would have been better but this is already used by another operator.

References:
https://discuss.lightbend.com/t/ideas-on-how-to-implement-a-latency-monitor-which-wraps-a-flow/5714/5

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,8 @@ libraryDependencies ++= Seq(
   "junit" % "junit" % "4.12" % Test, // Common Public License 1.0
   "com.novocode" % "junit-interface" % "0.11" % Test, // BSD-like
   "com.google.jimfs" % "jimfs" % "1.1" % Test, // ApacheV2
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @lostiniceland,

lostiniceland:
Can someone give a tip why Scala 2.12 is not working...

raboof:
Was it something like `scala.Seq` used to be `scala.collections.mutable.Seq` in 2.12 but now is `scala.collections.immutable.Seq` in 2.13, so importing the right one should help?

PR #188, opened 2021-04-09
AccumulateWhileUnchanged: stop pull() after onTimer()

ffettinger:
If `onTimer()` is called when `buffer` is not empty, it will `push()`
results but the `pull(in)` request will not be canceled. If additional
upstream elements arrive with a different `nextState` which would cause
`pushResults()` to be called again before the downstream indicates demand
via `onPull()`, then `nextElement` would be dropped.

Avoid pushing again after push() from onTimer() until onPull() is called
again to indicate downstream demand. This avoids dropping elements when
pushResults() is called but `out` is not available.

In case a similar issue occurs again, add a `require()` check which will
throw if nextElement would be discarded.

Note that the use of the `downstreamWaiting` flag is very similar to that of
`TwoBuffer` in the [Custom stream processing - Rate decoupled operators][custom-stream-rate-decoupled] example.

[custom-stream-rate-decoupled]: https://doc.akka.io/docs/akka/2.6.13/stream/stream-customize.html#rate-decoupled-operators

### DIFF
--- a/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
+++ b/src/main/scala/akka/stream/contrib/AccumulateWhileUnchanged.scala
@@ -68,6 +68,7 @@ final class AccumulateWhileUnchanged[Element, Property](propertyExtractor: Eleme
 
     private var currentState: Option[Property] = None
     private var nbElements: Int = 0
+    private var downstreamWaiting = false
     private val buffer = Vector.newBuilder[Element]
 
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @ffettinger,

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

johanandren (approved):

